PR #39, opened 2015-01-21
Allow centering a step over a specific element

jezstephens:
I think it's useful to be able to be able to center a step over a particular element, for example when you have a large element which occupies the whole screen which you need to explain.

This change lets you do this with `attachTo: '#widget center'` or `attachTo: {element: '#widget', on: 'center'}`.


b-ash:
No argument here. This needs to be rebased as well due to the last PR that was merged.

jezstephens:
Done!

PR #41, opened 2015-02-07
add 'before' trigger to Step

razor-1:
There are times when it's necessary to do something prior to the render() call in Step.show().  This adds a before trigger to Step.

### DIFF
--- a/shepherd.js
+++ b/shepherd.js
@@ -1604,6 +1604,7 @@ return this.Tether;
 
     Step.prototype.show = function() {
       var _this = this;
+      this.trigger('before');
       if (this.el == null) {
         this.render();
       }


rwwagner90:
I would very much appreciate it if this were merged!

zackbloom:
Added by https://github.com/HubSpot/shepherd/commit/241d63a4313dd81aef8d540293d1475be5f1c2b9, released in https://github.com/HubSpot/shepherd/releases/tag/v0.7.0.

rwwagner90:
@zackbloom would it be possible to implement skipping steps when the attachTo is not present now? I can do a hacky workaround with this before-show hook now, but if you guys are going to implement it, I could just wait. Is this being worked on?

PR #48, opened 2015-03-02
Add js file to main property

DanielApt:
I'm installing shepherd with NPM by listing it as a dependency in my package.json
```
"dependencies": {
"shepherd": "git://github.com/HubSpot/shepherd.git#58a37fa0263bf205fff7a64abb62216b1738e0c2",
},
```

Webpack and Browserify would require the main js file to be listed in the `main` property, instead it is listed in `spm.main`. I'm not using `spm`

For now I've decided to keep both, to not break `spm` support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
         "gulp-bower": "0.0.1",
         "gulp-wrap-umd": "*"
     },
+    "main": "shepherd.js",
     "spm": {
         "main": "shepherd.js"
     }


DanielApt:
Shouldn't have closed it, as most `package.json` files should use the `main` property. Feel free to merge.

DanielApt:
This is a near duplicate of a merged pull-request on [*tether*](https://github.com/HubSpot/tether/pull/68), so I'd imagine this should be fine to merge as well.

b-ash:
Sorry for the delay, and thanks for the update.

PR #51, opened 2015-03-16
updated error description to be more descriptive

juliankrispel:
This error message was causing me a bit of a headache because I never knew which element caused it.

Appending the selector helps a little. Hope that helps others too.


b-ash:
I'm happy to merge this if you don't mind rebasing with master. Also, you may consider either not defining `selector` or using it for the other two references to `opts.element`

juliankrispel:
Hi @b-ash. Ouch I just saw that you wanted a rebase. I just pulled from master basically, let me know if that's cool.

Also btw, having the bundled javascript in the git repo makes for a very noisy git history.
You could probably have a [preinstall hook](https://docs.npmjs.com/misc/scripts) in your package.json that way the project builds automatically when you install the package.

b-ash:
Thanks for the update. Keeping the compiled JS in the repo is indeed a bit of a pain, but it lowers the barrier to entry for folks wanting to toss it in their project quickly. We'll keep it in mind going forward if it becomes a real burden :smile_cat:

juliankrispel:
Thanks for merging. FYI ->
https://help.github.com/articles/creating-releases/

PR #56, opened 2015-04-02
Made before-show execute a callback

rwwagner90:
This allows before-show to fully finish executing before show is called, thus potentially fixing issues like attachTo not existing yet, etc.

### DIFF
--- a/coffee/shepherd.coffee
+++ b/coffee/shepherd.coffee
@@ -116,22 +116,22 @@ class Step extends Evented
     @tether = new Tether extend(tetherOpts, @options.tetherOptions)
 
   show: =>
-    @trigger 'before-show'
+    show =(=>
+      if not @el?
+        @render()
 
-    if not @el?
-      @render()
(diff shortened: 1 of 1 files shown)


rwwagner90:
This requires my Tether PR to be merged in first: https://github.com/HubSpot/tether/pull/74

b-ash:
I'm worried this is turning into a bit of a rabbit hole. Why not just allow an option when instantiating a `Step` that let's you define a promise for before show/hide. if it's there, all it and wait for the promise to resolve, if it's not, just execute the function now. You could do something like having `show` handle that logic and a `_show` that does the internal "showing of this step" bit. Perhaps it could relate to `hide` too, but I'm not sure it's as relevant in that situation.

rwwagner90:
@b-ash It really has been tough to wrap my head around. I'm open for doing it with any method that you see fit. The huge hurdle I have is the compiled JS is near impossible to read, and I don't really know CoffeeScript.

I'll try again.

PR #57, opened 2015-04-02
Added beforeShowPromise

rwwagner90:
You can now pass in a function that returns a promise that will be run before-show.

This works much better than the callback stuff I was trying to do. Thanks for the suggestion @b-ash :+1: . What do you think about this?


rwwagner90:
If you like this method, I can update the docs and add them to this PR as well. Please let me know.

coffee/shepherd.coffee
```diff
@@ -117,6 +117,13 @@ class Step extends Evented
     @tether = new Tether extend(tetherOpts, @options.tetherOptions)
 
   show: =>
+    if @options && @options.beforeShowPromise && typeof @options.beforeShowPromise is 'function'
```
b-ash: This is basically what I was thinking. A couple quick CoffeeScript tweaks:

- `@options` is guaranteed to be defined to at least an empty object `{}` thanks to [the constructor](https://github.com/HubSpot/shepherd/blob/master/coffee/shepherd.coffee#L45)
- The existential operator `?` does more than just loosely translate to `x !== null || x !== undefined`. When using it on a function call, say `@options.beforeShowPromise?()`, it will check that `typeof fun is 'function'`. CoffeeScript.org [has a nice repl](http://coffeescript.org/#try:x%3F.fun()%0A%0Ax.fun%3F()) you can play with.

One way to go about it would be to check that the result of `@options.beforeShowPromise?()` is defined (which you can use `?` for), and then either attach the promise or call the function. It may make things a bit clearer by defining the result anyways.

coffee/shepherd.coffee
```diff
@@ -117,6 +117,13 @@ class Step extends Evented
...
+      @options.beforeShowPromise().then(=>
+        @_show()
+      )
+    else
+      @_show()
+  _show: =>
```
b-ash: Super nitpicky - can we keep the spacing consistent, and have a carriage return between function definitions?

coffee/shepherd.coffee
```diff
@@ -117,6 +117,13 @@ class Step extends Evented
     @tether = new Tether extend(tetherOpts, @options.tetherOptions)
 
   show: =>
+    if @options && @options.beforeShowPromise && typeof @options.beforeShowPromise is 'function'
+      @options.beforeShowPromise().then(=>
```
b-ash: In this scenario, I'm probably more likely to do `.then(=> @_show())` on one line since spreading it out doesn't really do much besides making it harder to read. Since we don't have any args to pass along to `@_show` and `@_show` is defined with a fat arrow (`=>`), we can do `.then(@_show)`.

b-ash:
Besides those few stylistic comments, this is *much* better. I really like keeping this contained in here without needing to update Tether. If you can add doc updates to this as well, that would be fantastic. Thanks again for working through the comments!

rwwagner90:
I'll be glad to address those comments tomorrow. Do you also want the compiled js committed or just the CoffeeScript?

b-ash:
The compiled bit as well, thanks.

rwwagner90:
Sorry, should have been more specific. Just to clarify, just running ```gulp``` and committing everything? Like shepherd.js, shepherd-amd.js or whatever all those are?

As opposed to running ```gulp coffee```?

rwwagner90:
I got it down to just one line for you :). The compiled JS now looks atrocious, but at least the CS looks clean.

Let me know what you think.

Also updated the docs, and added documentation of before-show and before-hide, as well as being able to do things like:
```js
when: {
            show: function() {
                window.scrollTo(0, 0);
            }
          }
```

coffee/shepherd.coffee
```diff
@@ -117,6 +117,9 @@ class Step extends Evented
     @tether = new Tether extend(tetherOpts, @options.tetherOptions)
 
   show: =>
+      @options.beforeShowPromise?()?.then(=> @_show()) ? @_show()
```
b-ash: Nice. Mind simplifying `.then(=> @_show())` to `.then(@_show)`? In this case they're functionally the same.

rwwagner90: @b-ash They are not the same. They do not yield the same JS. Changing it to ```.then(@_show)``` removes the anonymous function from around it.

```js
Step.prototype.show = function() {
      var _base, _ref1, _ref2;
      return (_ref1 = typeof (_base = this.options).beforeShowPromise === "function" ? (_ref2 = _base.beforeShowPromise()) != null ? _ref2.then(this._show) : void 0 : void 0) != null ? _ref1 : this._show();
    };
```

vs this, where it's wrapped in ```function```
```js
Step.prototype.show = function() {
      var _base, _ref1, _ref2,
        _this = this;
      return (_ref1 = typeof (_base = this.options).beforeShowPromise === "function" ? (_ref2 = _base.beforeShowPromise()) != null ? _ref2.then(function() {
        return _this._show();
      }) : void 0 : void 0) != null ? _ref1 : this._show();
    };
```

The code does not work, if I switch it to the way you suggested.

b-ash: I tested on my end, and it works as we'd expect. And yes, it removes the anonymous wrapper, but it's unnecessary because `_show()` is a bound function (because of the fat arrow).

rwwagner90: Can you elaborate on how you tested it please? It definitely does not work on our end when I do the code your way.

b-ash:
```coffeescript
show: =>
  @options.beforeShowPromise?()?.then(@_show) ? @_show()

_show: =>
  ...
```

```coffeescript
tour = new Shepherd.Tour
tour.addStep 'start',
  title: 'Foo'
  text: 'bar'
  beforeShowPromise: ->
    p = Q.defer()
    setTimeout ->
      console.log('foo')
      p.resolve('foo')
    , 2000
    return p.promise

tour.start()
```

After starting the tour, it waits 2 seconds, logs 'foo', and then shows the tour. Commenting out the `beforeShowPromise` bit, it starts immediately. I happen to be using `Q` promises here, but it would work with anything that is thennable.

rwwagner90: After examining further, it appears that the issue for me is, ```this``` is not bound to the step as it should be. You should get errors on ```this.trigger('before-show')``` and such, because it expects this to be the step.

I'm 100% definitely seeing this behavior, so we need to bind ```this``` somehow. Is having one extra arrow detrimental to the code?

rwwagner90: I can create a jsfiddle or something, if you need me to reproduce the error for you, but it would be awesome if we could just merge it with the arrow :).

b-ash: Using the fat arrow (`=>`) to define `_show` adds a bind function call in the constructor to the instance when compiled. This makes it so that any function call to `_show` will in fact be the instance of the `Step` it was invoked in.

rwwagner90: I understand it's ***supposed*** to do that, but it's not binding to this as it should. I'll put in the time to reproduce the error for you, if you're not okay with merging it as is.

rwwagner90: I used a normal JS promise, and it still works. Something in my specific Ember app, makes it break, as the binding is messed up. Could we pretty please just merge it in with the fat arrow, as it makes it work for me? It should work for everyone. I think since it wraps it in another function, my bind doesn't get applied to it, and it still keeps the ```this``` reference.

rwwagner90: Could we keep the fat arrow please? Doing so wraps it in an anonymous function, and defines ```_this``` which keeps the context correct. Without it, my use case does not work. Is it detrimental to the codebase in some way if we keep it?

b-ash:
Thanks again for the update. We're excited to watch [ember-shepherd](http://rwwagner90.github.io/ember-shepherd/) as it comes along.

rwwagner90:
@b-ash Thanks for merging! Feel free to give any suggestions you may have for ember-shepherd :)

rwwagner90:
Does each merge become a new version number or do you update the version every so often? Definitely want to update my app with this version, but I didn't bump the bower or npm versions or anything because I didn't know how all that worked.

b-ash:
Depends, it's every update-ish :wink: I've gone ahead and updated those.

rwwagner90:
Cool, cool. Thanks!

PR #74, opened 2015-06-16
Avoid memory leaks due to double tether instance

opusonline:
@geekjuice Please have a look at my update suggestions.
I tried to explain the ideas in comments.
If you confirm, we can remove those comments.
Thanks, regards


geekjuice:
Awesome! I'll take a look sometime to make sure it doesn't break anything. LGTM though :+1:

PR #75, opened 2015-07-02
 Add support for all Tether attachment positions

JSMike:
This will resolve issue 67 that was raised.

If attachTo.on is found in ATTACHMENT object then defaults to old functionality
If attachTo.on is an empty string or not defined then defaults to ATTACHMENT.right
If attachTo.on is defined and not found in ATTACHMENT then passes string as tether attachment


geekjuice:
Awesome! I'll take a look and if everything looks good, have it merged in right away. Thanks! :smiley:

geekjuice:
Sorry for the delay. The solution is close, but the issue is that `ATTACHMENT[position]` will attach to the opposite of `position` e.g. `bottom -> top center` whereas the changes here will attach the shepherd step at `attachTo.on`. I remember this wasn't as trivial as I hoped when I took an initial swing at it too. Let me know if you wanna take another look. :+1:

JSMike:
Ok, but this solution lets the default functionality stand, while allowing for customization, assuming if someone was deliberately entering two words that work for tether that they could set the correct "opposite direction". If you want it to flip the words so it ends up in the opposite position I can look into that, but it might get confusing.

geekjuice:
Yea, I think the latter (library flips the words) makes more sense since it's more aligned with the current implementation. I think the current solution can just be expanded to lookup opposite positions from a constant map. For example:

```javascript
const FLIPPED_ATTACHMENT = {
  'top center': 'bottom center',
  'left middle': 'right middle',
   ...
};

let attachment = ATTACHMENT.right;
if (opts.on in ATTACHMENT) {
  attachment = ATTACHMENT[opts.on];
} else if (opts.on in FLIPPED_ATTACHMENT) {
  attachment = FLIPPED_ATTACHMENT[opts.on]
}
```

There are a few (probably better) ways since the above is just an example.

JSMike:
It might be easier to just add all possible combinations to ATTACHMENT

geekjuice:
Yup, I agree :sweat_smile:

JSMike:
Updated Attachment locations.
Updated logic for default location

JSMike:
@geekjuice let me know if i missed anything.

geekjuice:
Awesome. Thanks again for the updates.

I'll try to find some time today to take a look and make sure everything works as expected :+1:

geekjuice:
So this is looking much better. There's actually a few more permutations of attachments that seem missing e.g. `left top`. The [tooltip demo](http://github.hubspot.com/tooltip/docs/welcome/) site gives you all of the possible attachments.

However, another issue I noticed is that the shepherd styles with arrows don't seem to take into account the new attachments (since they didn't exist until this PR), so the arrows are misplaced. 

Let me know if this is still a undertaking you want to tackle alone, otherwise, I'll be happy to help out.

shaneparsons:
+1 – It would be awesome to have this functionality!

alecortega:
+1, this would be super helpful

okbel:
+1

geoffdaigle:
+1

geoffdaigle:
I'm not really sure why this still isn't supported, so I'm going to make this happen - @geekjuice @zackbloom stop me if I'm missing something

geekjuice:
Thanks for reviewing this PR. If it works, I say :rocket:. I've been _really_ behind on reviewing PRs :disappointed:

zackbloom:
Keep in mind that as @geekjuice mentioned, I don't believe the arrow styles are written to support these other attachment options, so they should be at the very least hidden if an unsupported attachment is used.

rwwagner90:
I believe this is obsolete, since #106 was merged

PR #78, opened 2015-08-24
Added showOn options parameter function

mtgibbs:
We have conditionally shown controls on our site and didn't want to have to make individual tours for different user types.  Please take a look at this proposed update inspired by Issue 65.

https://github.com/HubSpot/shepherd/issues/65

JsFiddle showing it in action:

http://jsfiddle.net/mattgibbs/f8n0gLrL/4/


src/js/shepherd.js
```diff
@@ -527,13 +528,19 @@ class Tour extends Evented {
     }
 
     if (next) {
-      this.trigger('show', {
-        step: next,
-        previous: this.currentStep
-      });
+      if (typeof next.options.showOn !== 'undefined' && !next.options.showOn()) {
+        const index = this.steps.indexOf(next);
+        var nextIndex = forward ? index + 1 : index - 1;
```
geekjuice: One minor detail. `var` here should be `const`.

geekjuice:
Other than the small comment above, I think you can go ahead and minor version bump this to `v1.2.0` since `showOn` is a new feature. To do that, just run `gulp version:minor && gulp build` to have all the `.json` files and distribution files updated.

Once that's done, I'll go ahead and merge this.

Thanks again for being patience while I was on hiatus :+1:

mtgibbs:
Great!  Thanks for reviewing it.

I've committed your feedback, bumped the version, and committed the install output.

I really enjoy these tours and if I run into something else I'll send it your way.

geekjuice:
:tada: :+1:

dmackerman:
Is there no documentation for `showOn`?

mtgibbs:
I guess not... :(

I didn't know where to submit the changes to the documentation.  But what is it you want to know about it?  Basically you have to define the showOn option as a function that returns a boolean.  If it is defined, when the step is stepped to it will execute the function and look at the return value to decide whether or not the step should be shown.  If it shouldn't be, it will skip to the next one.

PR #88, opened 2015-11-29
Patch fix to return the Step object in addStep method (Issue #76)

itamar:
A quick patch fix to resolve issue #76 - `addStep does not return the Step instance`


rwwagner90:
This was fixed in another commit.

PR #95, opened 2016-02-04
Fix retethering issue when anchor refs are removed 

geoffdaigle:
@zackbloom 

I am working on a hubspot app that has a very dynamic, multi-page onboarding flow. The problem happens _after_ the user finishes a flow and then decides to start it over. This situation might apply more broadly as well.

## Current behavior (attaches to 0,0)
tether-problem-shepherd

## Fixed behavior (attaches to element)
tether-problem-shepherd-fixed

The issue comes from the fact that the user options `opts.element` reference is overwritten when setting the element reference using `document.getQuerySelector` and stays in memory. Unless the steps are explicitly removed and added again (which, in practice, is not intuitive) the elements are never relieved of their reference and thus will always re-attach to the historical first reference that has no presence on the page.

~~(Additionally, a deep `extend` would have worked nicely here but it seems like that doesn't happen in Tether utils).~~ Nevermind that doesn't apply here


src/js/shepherd.js
```diff
@@ -177,18 +176,23 @@ class Step extends Evented {
...
 
   getAttachTo() {
+    let returnOpts = {};
     let opts = parsePosition(this.options.attachTo) || {};
-    const selector = opts.element;
-
-    if (typeof selector === 'string') {
-      opts.element = document.querySelector(selector);
+    returnOpts.on = opts.on;
+    if (opts.offset !== 'undefined') {
```
zackbloom: `opts.offset` really ends up the string `'undefined'`? That doesn't seem right.

geoffdaigle: oh good eye, i meant `typeof` here

dist/js/shepherd.js
```diff
@@ -207,18 +206,23 @@ var Step = (function (_Evented) {
...
+      var returnOpts = {};
       var opts = parsePosition(this.options.attachTo) || {};
-      var selector = opts.element;
-
-      if (typeof selector === 'string') {
-        opts.element = document.querySelector(selector);
+      returnOpts.on = opts.on;
+      if (typeof opts.offset !== 'undefined') {
+        returnOpts.offset = opts.offset;
+      }
```
zackbloom: I think you can just do `returnOpts = extend({}, opts)`

dist/js/shepherd.js
```diff
@@ -207,18 +206,23 @@ var Step = (function (_Evented) {
...
 
-        if (!opts.element) {
-          throw new Error('The element for this Shepherd step was not found ' + selector);
+      if (typeof opts.element === 'string') {
+        // Can't override the element in user opts reference because we can't
+        // guarantee that the element will exist in the future.
+        returnOpts.element = document.querySelector(opts.element);
+        if (!returnOpts.element) {
+          throw new Error('The element for this Shepherd step was not found ' + opts.element);
```
zackbloom: I'm a little concerned that throwing an error may be more aggressive than is necessary.  It might be better to use `console.error` here, to not prevent other js on the page from executing properly.

dist/js/shepherd.js
geoffdaigle: You're right - I got this confused with something else, it's only a single level of props so i wouldn't need deep copy anyway.

geoffdaigle:
@zackbloom anything else need tweaking before this is merged?

marcneuwirth:
:+1:

geoffdaigle:
@zackbloom im going to merge this

timmfin:
Publishing as `1.6.0` shortly...

PR #99, opened 2016-02-09
Change UTF-8 Character to escaped Unicode

Beldur:
This change prevents accidental encoding conversion errors of the cancel icon.

### DIFF
--- a/src/js/shepherd.js
+++ b/src/js/shepherd.js
@@ -328,7 +328,7 @@ class Step extends Evented {
     }
 
     if (this.options.showCancelLink) {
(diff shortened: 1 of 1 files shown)


geoffdaigle:
:+1:

geoffdaigle:
@Beldur on closer look at this, the gulp file will encode the `\u2715` and leave it as `✕` when built and minified. I'm going to revert this PR for now, but I'll add it back if you or I manage to change the build config to leave it as-is.

(cc @geekjuice Thoughts on this? You may know where that happens off the top of your head)

PR #100, opened 2016-02-09
More robust button config

geoffdaigle:
Fix for https://github.com/HubSpot/shepherd/issues/94

#### Bumps from v1.6.2 to v1.7.0

Besides fixing the problem where `{buttons: false}` doesn't stop the default button from rendering, this PR makes the button API more flexible:

__When 'buttons' equals:__
`[ {buttonConfig} ... ]` --> Regular behavior of displaying custom buttons
`undefined/true` --> Default "next" button with `btn` class
`{singleButtonConfig}` --> `[ {singleButtonConfig} ]`
`{} / [] / false / null` --> No buttons and no footer

In addition, I moved `undefined` and `object` strings into a few methods that handle the same thing in a safer way.

@marcneuwirth @zackbloom


src/js/shepherd.js
```diff
@@ -11,6 +11,9 @@ const {
 
 let Shepherd = new Evented;
 
+const UNDEFINED = 'undefined';
```
mattrheault: If the optimization here is to not have to write ```'undefined'``` everywhere, then we're better off creating a helper ```isUndefined()``` method to this check for us. This will reduce redundancy even further.

```javascript
const isUndefined = (obj) => {
  return typeof obj === 'undefined';
};
```

dist/js/shepherd.js
```diff
@@ -157,12 +173,33 @@ var Step = (function (_Evented) {
         }
       }
 
-      if (!this.options.buttons) {
+      // Button configuration
+
+      var buttonsJson = JSON.stringify(this.options.buttons);
```
geoffdaigle: Note to self - this conditional stuff might be able to be abstracted to other places, but this is the only area that uses "python"-y truthy/falsey rules right now.

dist/js/shepherd.js
```diff
@@ -157,12 +173,33 @@ var Step = (function (_Evented) {
...
           text: 'Next',
-          action: this.tour.next
+          action: this.tour.next,
+          classes: 'btn'
         }];
-      }
+
+        // Can pass in an object which will assume asingle button
+      } else if (!buttonsAreEmpty && buttonsAreObject) {
+          this.options.buttons = [this.options.buttons];
```
geoffdaigle: Now silently accepts an object and keeps the later `map

PR #106, opened 2016-03-19
All tether position and arrow fix 

geoffdaigle:
@geekjuice @zackbloom 

__This will bump the shepherd version to 1.8.0__

Instead of bothering with merge conflicts, I decided to port https://github.com/HubSpot/shepherd/pull/75 into this branch and test the arrow styles here.
I added a utility to the shepherd theme which hides arrows on diagonal (corner-to-corner) tether attachments.

One problem that came up is that gulp won't let me minify `shepherd.js`, throwing this error:

```
message: '/Users/gdaigle/repos/shepherd/dist/js/shepherd.js: Object Tether has no method \'push\'',
fileName: '/Users/gdaigle/repos/shepherd/dist/js/shepherd.js',
lineNumber: undefined,
```
(I had started by doing a fresh install of npm and bower dependencies, for context.)

You'll notice that the unminified version is here, where it should be minified. Any ideas on how to get that working? Otherwise I think this is looking good.


chicoxyzzy:
hope to see this merged soon :+1:

geoffdaigle:
@zackbloom @alecortega @mattrheault @geekjuice 

Do y'all have a minute to check out this build? Gonna give it another try but it would be cool to get this merged soon. Don't want my local env to be a blocker for a reasonable update.

alecortega:
@geoffdaigle I'm getting a very similar error when I try to build. Gonna investigate this issue, there's a number of other problems on initial install of this project too that I'll try to make PR's for as I'm working through this and you can just rebase on top of that. Right now it seems as though this issue is blocking all development on Shepherd. 

```
stream.js:74
      throw er; // Unhandled stream error in pipe.
      ^
TypeError: options.except.push is not a function
  at AST_Node.eval [as mangle_names] (<anonymous>:3528:20)
```

alecortega:
@geoffdaigle @timmfin 
Alright, it seems as though the cause of this break is in this commit: https://github.com/HubSpot/shepherd/commit/d5fe2aeb0e95c59d084e0134abcc302aac6445fb. 

There's a mismatch between the function name Shepherd is looking for and Tether itself since we're only mangling Shepherd and not Tether. 

Option 1: We don't mangle either which would result in a slightly larger file size. 
Option 2: We mangle both, but it seems @timmfin had a specific reason for not doing so on Tether. 

More info would be great surrounding this. But, I feel like the slightly larger file size is better than a blocker on all development on this project, at least until we get this resolved. Thoughts?

geekjuice:
The `mangle.except` option in `gulp-uglify/uglify-js` takes an `Array`. The fix should be as easy as:

```js
// gulpfile.js
.pipe(uglify({
  mangle: {
    except: ['Tether']
  }
}))
```

Credit to [gulp-uglify#67](https://github.com/terinjokes/gulp-uglify/issues/67)

@geoffdaigle @alecortega @timmfin

geekjuice:
@timmfin Correct me if I'm wrong, but because we are using the UMD wrapper, mangling `Tether` should be safe since it doesn't matter what the variable name is within the wrapper since the arguments that pass the instance will always remain correct.

timmfin:
It took a while, but found some context in an old internal chat thread. Looks like I was quickly trying to fix an issue with the [Shepherd docs](http://github.hubspot.com/shepherd/). But I'm skeptical that my fix was the correct thing to do (despite my memory that I did solve a problem).

So, my guess is that we should go ahead and remove the mangling exception and if something comes up again that we need to re-fix, so be it (and if it does, I suspect there is a better fix than my quick hack).

PR #118, opened 2016-05-23
Change X to HTML Equivalent: &#10005;

zeg-io:
The problem with having the raw UTF8 character in there rather than the HTML equivalent is that it breaks compatibility under certain circumstances.  In my opinion the few extra characters are well worth the compatibility gains.  The difference is six bytes.  This will resolve the issue I previously posted on the odd characters returning rather then the multiplication character (or X).

I see that previously there was an escaped unicode character, I didn't try it with that, but I'm guessing it was added for similar reasons.


yairEO:
I prefer to always use `&times;`

zeg-io:
That would be fine as well, right now it's broken for me though, arg.  @HubSpot please fix your own way or merge this

djsegal:
+ why is the character that gets used not an option like `text`?
+ i would like to use `&times;` as well

rwwagner90:
There is another PR out that removes the `x` and uses css to add it in, so this should be easily overridable https://github.com/shipshapecode/shepherd/pull/148

PR #121, opened 2016-06-03
check for null returned when an element for a step is not found

BaTbAbYDeV:
If getAttachTo() fails to find an element passed over as a string, it sets that element to null.  This bypasses the check in setupTether which only checks if opts.element is undefined and the step does not get shown leaving the tour in an undetermined state.

This fix checks for null and presents the step in the middle of the viewport as expected.


rwwagner90:
I think this is the intended behavior. You have to not specify `attachTo` at all, meaning it is undefined, to get it to appear in the center of the screen. If the element is not found, it should be an error, not just shown in the middle.

PR #126, opened 2016-06-25
Adds Bower install to postinstall

alecortega:
Documentation states that both node_modules and bower_components are install when running npm install which wasn't the case. Added bower as a local dependency and bower install to the postinstall script to fix this.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "tether-shepherd",
-  "version": "1.7.0",
+  "version": "1.7.1",
   "homepage": "https://github.com/HubSpot/shepherd",
   "authors": [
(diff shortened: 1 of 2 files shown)


geekjuice:
I believe dependencies on `bower` were removed for the `tether` based projects. Mainly because having two registries to depends on seemed unnecessary and required users to install `bower`. I would be more inclined to update the documentation to omit `bower` and replace `bower` usage with `npm/node_modules`.

alecortega:
@geekjuice Sounds like a better option to me, will update the PR to reflect those changes.

darrynten:
Any reason you need bower? Is npm not suitable to your needs?

rwwagner90:
I don't think we need bower anymore, so closing this PR.

PR #147, opened 2017-01-06
Ability to set classPrefix as default value

bm2u:
- for tether and shepherd-elements at once
- pathfinder actions (missing semicolons)
- makes custom styling much more easier

### DIFF
--- a/src/js/shepherd.js
+++ b/src/js/shepherd.js
@@ -119,6 +119,25 @@ function parseShorthand (obj, props) {
   return out;
 }
 
+function getDefaultsKey( obj, key){
+
+    if (isObjectLoose(obj)) {
+
+        if ( obj.hasOwnProperty( 'defaults' ) ) {
+
+            const defaults = obj.defaults;
+
+            if ( defaults.hasOwnProperty( key ) ) {
+                return defaults[ key ];
+            }
+
+            return 'shepherd';
+        }
+    }
+
+    return 'shepherd';
+};
+
 class Step extends Evented {
 
   constructor(tour, options) {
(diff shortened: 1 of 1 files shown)


rwwagner90:
@bm2u couldn't you just style based on the `shepherd-` classes?

rwwagner90:
@bm2u we've now replaced tether with popper.js, so this eliminates some of this I think. Do we still need this?

bm2u:
@rwwagner90 , I am still convinced that we need a possibility to set a different prefix (option) than e.g. "popper" or "shepherd" via JS. 

Look at the following two files:
https://github.com/shipshapecode/shepherd/blob/master/src/css/_variables.scss
https://github.com/shipshapecode/shepherd/blob/master/src/css/helpers/_popper.scss

If you use Shepherd/Popper in combination with SCSS, you have the opportunity to set theme prefixes. These prefixes are a part of the final CSS result. Without the possibility to pass the prefix via Javascript, the tour would remain unstyled as a reason of the hardcoded default JS values "popper" and "sheperhed".

rwwagner90:
I'm going to close for now. I really appreciate the PR and the discussion @bm2u. Please feel free to come back and open another PR if you decide on reasoning behind why we should do this and how it would work with the built in shepherd themes. In the meantime, I think you can use `classes` and custom CSS to solve this.

cbou:
Well I think it not a bad thing. Having MyPopover-text instead shepherd-text looks much better integrated into your app. Currently our app we do already have popover we could reuse for the tour. Integrating Shepherd would be much easier with this feature.

Tippyjs allow to override the prefix class.

rwwagner90:
@cbou I'm not sure I understand. We use tippy under the hood in Shepherd and allow passing all tippy options to it. What is it you would like here and would you like to open another issue to discuss it perhaps?

PR #148, opened 2017-01-06
Close button as HTML entity

bm2u:
- adjust styling for available themes
- removed useless properties


rwwagner90:
Thanks so much!

PR #149, opened 2017-01-11
Add Repo to package.json

bm2u:
- added property with valid link to public repo to package.json


rwwagner90:
Now that this has been transferred to us, these values are out of date. I need to get the demos and docs deployed again, and we can update these.

rwwagner90:
Thanks for the PR!

PR #158, opened 2017-04-22
Added support to open Shepherd inside <dialog> elements with the correct offset.

MarkusBordihn:
Fixed https://github.com/HubSpot/shepherd/issues/157

Note: Will only work together with the latest version of tether, but will also not breaking anything if an older version is used.


rwwagner90:
Hi @MarkusBordihn thanks for the PR! We recently merged in a refactor where we are now using popper.js instead of tether. Could you please give the latest beta a try and see if this is still an issue? If so, do you mind rebasing? We no longer check in the dist files, so you can remove all of those.

MarkusBordihn:
@rwwagner90 
Thanks for the update. I can confirm that this is still an issue even with the latest version.
I will check if I could provide an similar fix for the latest version.

PR #160, opened 2017-05-12
Add button option "defaultButtonCls"

arshtepe:
This option allow remove  .shepherd-button  class from button.
[defaultButtonCls=undefined]


rwwagner90:
Unfortunately, I am not really in favor of removing the classes from the buttons, at this time. If the idea was just to style differently, I would recommend just using custom CSS.

PR #176, opened 2017-11-27
Change positioning engine from Tether to Popper.js

lancegliser:
The existing tether project has been discontinued as noted on their page. Many projects have begun to move away from tether, to Popper as suggested by Hubspot. My own need was because of Bootstrap 4's migration. I wanted Shepherd, but didn't want to include Tether just for it.

I expect this will have some bugs and use cases missing, but it's a functional start to begin talking about how options, themes, etc might change in a 2.x popper branch.

Conceptually, little has changed:
- The arrow css depends on a dom element instead of a :before as popper targets it differently

Outstanding issues:
- Orphan steps using positionFixed in popper's upcoming release


lancegliser:
Associating this with #163

bower.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "tether-shepherd",
+  "name": "tether-popper",
```
rwwagner90: Shouldn't this be `popper-shepherd`?

dist/css/shepherd-theme-arrows-fix.css
```diff
@@ -1,5 +0,0 @@
-.shepherd-element-attached-bottom.shepherd-element-attached-right.shepherd-target-attached-top.shepherd-target-attached-left .shepherd-content:before,
```
rwwagner90: Is the arrow fix no longer needed?

rwwagner90:
@lancegliser I tried this code out and it seems to work pretty well. How should we implement the orphaned steps stuff? Is that supported by popper now?

rwwagner90:
I'm going to merge this into the 2.0-alpha branch for now, until I can get the orphaned elements to work.

lancegliser:
@rwwagner90 Sorry I've been very bogged down with things. Happy to see this was a pleasing start to 2.x for you. I should have some time in the next month to attempt the orphan code as well if that's soon enough for you.

rwwagner90:
@lancegliser I took a stab at it, but was unable to get offsets to work, so the orphaned element is not in the center of the screen. It may be easier to just copy the centered fixed logic from tether and not use popper for this case.

PR #181, opened 2018-04-05
Attempting to fix uncaught exception caused by non-existing element f…

IvanKarpan:
…or shepherd attachment

Hi,

There is a case when document.querySelector can throw an exception which was not handled by the library and therefore caused me some troubles (beforeShowPromise didn't fire and my perfectionism  suffered like a lot).

As per documentation on document.querySelector: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

Please kindly accept this pull request to right all wrongs.


rwwagner90:
@IvanKarpan can you please elaborate on this? How can `document.querySelector` throw an exception?

IvanKarpan:
It happens if the selector used is malformed. I ran into this scenario because I was working on a tour guide builder of sorts where users can populate an array of steps by providing a selector and a string for each item. And the problem surfaced when they input an invalid selector which caused an exception I could not handle. Hence the pull request. :)

rwwagner90:
@IvanKarpan but `document.querySelector('fdsafdsa')` works fine and returns `null` it does not throw an error. Can you be more specific about how there was an error thrown here please?

IvanKarpan:
@rwwagner90 try doing `document.querySelector('')`, I mean in a scenario where it's an input from the user you can expect anything. Even if it's the developer tinkering with the library it should not shoot out exceptions for an empty string selector. The pull request I suggested catches this situation, the variable does not get populated and the library processes this as if it hasn't found the element on the page.

rwwagner90:
Thank you for clarifying. I just thought `document.querySelector` handled its own errors. Now that I know the case in which errors are thrown, I am in support of this.

rwwagner90:
@IvanKarpan I just merged the conversion to popper, could you please rebase?

rwwagner90:
I fixed the conflicts in another branch and merged. Thanks so much for the PR!

IvanKarpan:
Sorry, missed your request, thx a lot for handling the conflicts and accepting the PR!

PR #204, opened 2018-08-07
improved themeing

bm2u:
improved abilities to create your own themes by changing corresponding themeing vars


bm2u:
@rwwagner90 Don't know why the old "classPrefix-Stories" were mentioned above. There is nothing about that inside this PR. Feel free to review.

rwwagner90:
Hi @bm2u, thanks for the PR! How would someone not using SCSS do themes here? The idea is you can just include theme files and get the themes automatically. We could perhaps use CSS variables, instead of SCSS, but I'm not sure on how to best handle that. 

We're in the middle of a big refactor right now, to use Webpack and further simplify the code, so we may want to revisit this after we get all that merged in.

bm2u:
Hi @rwwagner90 , you're welcome. I agree with you when you say we put the PR aside for now. However, nothing will change for CSS users. Default theme-files are still included. Only the creation of the themes will be easier (for you and actually SASS users). 

Feel free to adapt the default themes to your needs and deliver the dist-folder as before. CSS users need to work with custom CSS classes, or overwrite the values (as before).

BTW: If your are in need of any support about gulp oder webpack, do not hesitate to contact me.

rwwagner90:
Hi @bm2u we just merged in the webpack changes. I'd like to get some tests written as the next step now, since we have karma and mocha up and running. If you'd like to help write some tests, we'd love the help! 

Once we have good tests in place, we can definitely make sweeping changes like this and be confident everything still functions mostly the same.

rwwagner90:
@bm2u could you give me a quick TL;DR of what is going on in this PR? The main questions I have are:

- Will there be breaking changes for current CSS users or will everything work the same?
- For SCSS users, can we document the process of how they would import and override variables?

bm2u:
The process itself does not change anything for CSS users. I made some color adjustments to the themes, but we could certainly reduce them to the original look. Finally, with the advanced theming option, we have many adjustment screws to turn on.

The process for SCSS users is documented (see README).

rwwagner90:
@bm2u thanks for writing those docs! Can we put them in `index.md` instead of `README.md`? `index.md` is what we use for the API docs.

Also, could we limit the scope of this PR to just allowing theming more easily, and not changing the default styles? I'd like to make sure this generates the same styles as before, then we can think about changing the default styles later.

bm2u:
@rwwagner90 my pleasure. Previous appearance of delivered themes should be restored. Some small but helpful additions in step.js (add css classes to prevent the styling of a particular html element). It seems to me that some of the themes seem to be visualy identical. Nevertheless, I have tried to get identical results at all. Index.md contains helpful details about using themes.

index.md
```diff
@@ -261,6 +261,67 @@ mediator.on('user-create', () => {});
...
+```javascript
+let tour = new Shepherd.Tour({
+  defaults: {
+    classes: 'shepherd-theme-custom'
+  }
+});
+```
+
+### Custom Themeing
+We use [SASS](https://sass-lang.com/) as pre-processor for CSS. In connection with SASS there are extensive possibilities to generate CSS. For example, SASS can calculate or increase the saturation of color values. In addition, variables can be defined (similar to a scripting language), which ultimately end up as values in the CSS result. We make use of these extended possibilities by extracting themeing-relevant values as variables (__variables.scss_). This makes it easy to individualise colors and shapes.
```
chuckcarpenter: `(__variables.scss_)` should be `(__variables.scss__)` so that all following content isn't bold

PR #205, opened 2018-08-08
multi-platform clean script

bm2u:
Usage of del-cli to run clean script on windows as well


bm2u:
Assuming your git install is in the default path, just add C:\Program Files\Git\usr\bin to your %PATH% and the issue will disappear (same with copy/cp)

PR #215, opened 2018-08-22
Testing for Evented module

chuckcarpenter:
this branch adds some tests for the base class and increases code coverage

```
TOTAL: 21 SUCCESS
-------------|----------|----------|----------|----------|-------------------|
File         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------|----------|----------|----------|----------|-------------------|
All files    |    72.01 |    59.19 |       64 |    72.78 |                   |
 evented.js  |    84.85 |    80.77 |      100 |    84.85 |    30,31,32,33,35 |
 shepherd.js |      100 |      100 |      100 |      100 |                   |
 step.js     |    66.67 |    55.12 |       40 |    68.02 |... 84,385,400,401 |
 tour.js     |    73.81 |       50 |      100 |    73.49 |... 61,165,166,167 |
 utils.js    |    90.48 |     87.5 |      100 |    90.48 |             34,52 |
-------------|----------|----------|----------|----------|-------------------|
```

### DIFF
--- a/src/js/evented.js
+++ b/src/js/evented.js
@@ -21,7 +21,7 @@ export class Evented {
 
   off(event, handler) {
     if (typeof this.bindings === 'undefined' || typeof this.bindings[event] === 'undefined') {
-      return;
+      return false;
     }
 
     if (typeof handler === 'undefined') {
--- a/test/test.evented.js
+++ b/test/test.evented.js
@@ -2,7 +2,7 @@
 import { assert } from 'chai';
 import { Evented } from '../src/js/evented';
 
-describe('Evented', function() { //eslint-disable-line mocha/no-exclusive-tests
+describe('Evented', function() {
   const testEvent = new Evented();
   let testOnTriggered = false;
   describe('on()', function(){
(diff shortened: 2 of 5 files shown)


src/js/evented.js
```diff
@@ -21,7 +21,7 @@ export class Evented {
 
   off(event, handler) {
     if (typeof this.bindings === 'undefined' || typeof this.bindings[event] === 'undefined') {
-      return;
+      return false;
```
rwwagner90: Why did we change to return false here? Not saying it's wrong, just want to understand what it does for us.

chuckcarpenter: clearer response for tests

rwwagner90: I wonder if maybe we should skip the return part, and just wrap the whole method in 
```js
if (!_.isUndefined(this.bindings) && !_.isUndefined(this.bindings[event])) {
```

rwwagner90: Perhaps we should save that for a later refactor though, since we might replace evented anyway

PR #216, opened 2018-08-23
Add some tour tests

rwwagner90:


src/js/tour.js
```diff
@@ -38,17 +38,27 @@ export class Tour extends Evented {
...
+   * When arg2 is defined, arg1 can either be a string or number, to use for the `id` for the step
+   * When arg2 is undefined, arg1 is either an object containing step options or a Step instance
+   * @param {Object|Step} arg2 An object containing step options or a Step instance
+   * @returns {Step} The newly added step
+   */
+  addStep(arg1, arg2) {
+    let name, step;
+
+    // If we just have one argument, we can assume it is an object of step options, with an id
+    if (_.isUndefined(arg2)) {
```
chuckcarpenter: what's the benefit of renaming the args?

rwwagner90: They were not truly `name` and `step`. Really, this is behaving as an overloaded method. If there is just one arg, there is no `name`, but name is the first param. It confused me with the naming, so I renamed them.

PR #217, opened 2018-08-23
Reduce Evented complexity

rwwagner90:


src/js/evented.js
```diff
@@ -1,15 +1,13 @@
...
-    // TODO: do we need this empty constructor?
-  }
+import _ from 'lodash';
 
+export class Evented {
   on(event, handler, ctx) {
     const once = arguments.length <= 3 || arguments[3] === undefined ? false : arguments[3];
 
-    if (typeof this.bindings === 'undefined') {
+    if (_.isUndefined(this.bindings)) {
```
chuckcarpenter: Think the original intent is to check if it's an object, perhaps that one instead?

src/js/evented.js
```diff
@@ -20,51 +18,36 @@ export class Evented {
   }
 
   off(event, handler) {
-    if (typeof this.bindings === 'undefined' || typeof this.bindings[event] === 'undefined') {
+    if (_.isUndefined(this.bindings) || _.isUndefined(this.bindings[event])) {
```
chuckcarpenter: same for this first statement

src/js/evented.js
```diff
@@ -20,51 +18,36 @@ export class Evented {
...
-      const _len = arguments.length;
-      const args = Array(_len > 1 ? _len - 1 : 0);
-      let i = 0;
+    if (!_.isUndefined(this.bindings) && this.bindings[event]) {
+      const args = _.drop(arguments);
 
-      for (let _key = 1; _key < _len; _key++) {
-        args[_key - 1] = arguments[_key];
-      }
+      this.bindings[event].forEach((binding, index) => {
```
chuckcarpenter: so much easier to read.

PR #222, opened 2018-08-25
Move bind methods to their own file

rwwagner90:


src/js/step.js
```diff
@@ -77,14 +87,14 @@ export class Step extends Evented {
     const text = createFromHTML('<div class="shepherd-text"></div>');
     let paragraphs = this.options.text;
 
-    if (typeof paragraphs === 'function') {
+    if (_.isFunction(paragraphs)) {
```
chuckcarpenter: much cleaner

PR #226, opened 2018-08-29
Implement ESDoc

rwwagner90:


src/js/step.js
```diff
@@ -322,32 +310,91 @@ export class Step extends Evented {
...
+    _.forOwn(when, (handler, event) => {
+      this.on(event, handler, this);
+    });
+
+    this._setUpButtons();
+  }
+
+  /**
+   * Wraps `_show` and ensures `beforeShowPromise` resolves before calling show
+   * @return {*}
```
chuckcarpenter: does esdoc not support `{Promise}` here as the returned value?

src/js/step.js
```diff
@@ -322,32 +310,91 @@ export class Step extends Evented {
...
     }
+    this._show();
+  }
 
-    if (this.popper) {
-      this.popper.destroy();
+  /**
+   * Determines button options prior to rendering
+   *
+   * @private
```
chuckcarpenter: still should have `@return` for readability

src/js/tour.js
```diff
@@ -177,6 +142,51 @@ export class Tour extends Evented {
...
+
+      // If we have steps left, show the first one, otherwise just cancel the tour
+      this.steps.length ? this.show(0) : this.cancel();
+    }
+  }
+
+  /**
+   * Setup a new step object
+   * @param {Object} stepOptions The object describing the options for the step
+   * @param {String|Number} name The string or number to use as the `id` for the step
```
chuckcarpenter: seems like type should be limited here

rwwagner90: Both numbers and strings are supported

src/js/step.js
rwwagner90: It does, we don't always return a promise though, if `beforeShowPromise` is not a function.

PR #228, opened 2018-08-29
[Bug] - fix cancel on any step with a title

chuckcarpenter:
This branch changes how the title is added to the step nodes, since using innerHTML will remove any event listeners which where attached for the cancel action.

A couple new tests were added to improve coverage as well.


package.json
```diff
@@ -50,7 +50,8 @@
   "dependencies": {
     "element-matches": "^0.1.2",
     "lodash": "^4.17.10",
-    "popper.js": "^1.14.3"
+    "popper.js": "^1.14.3",
+    "sinon": "^6.1.5"
```
rwwagner90: Sinon should be a devDep

src/js/step.js
```diff
@@ -138,6 +137,14 @@ export class Step extends Evented {
     const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id="${this.options.idAttribute}"}>`);
     const header = document.createElement('header');
 
+    if (this.options.title) {
```
rwwagner90: Looks good! 👍

PR #233, opened 2018-08-30
add onCapture setting to bubble events such as blur for tour elements…

chuckcarpenter:
… that may not exist

### DIFF
--- a/src/js/bind.js
+++ b/src/js/bind.js
@@ -26,14 +26,14 @@ export function bindAdvance() {
   const handler = _setupAdvanceOnHandler.call(this, selector);
 
   // TODO: this should also bind/unbind on show/hide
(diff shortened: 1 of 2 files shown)


src/js/bind.js
```diff
@@ -26,14 +26,14 @@ export function bindAdvance() {
   const handler = _setupAdvanceOnHandler.call(this, selector);
 
   // TODO: this should also bind/unbind on show/hide
-  if (!isUndefined(selector)) {
-    const el = document.querySelector(selector);
+  const el = document.querySelector(selector);
+  if (!isUndefined(selector) && el) {
```
chuckcarpenter: event handlers lost on any selector that is null at the time, since the step might not be in the DOM yet.

PR #239, opened 2018-09-06
Syntax errors and ParentNode.prepend not supported in IE

### LINKED ISSUE #238: Build syntax errors and element.prepend function not supported (Internet Explorer)
Hello,
Latest releases(post 2.0.0.beta-21) seem to have some build syntax errors which prevent the lib to loaded in **IE**. Also, once these errors are corrected and the lib has loaded, a call to ParentNode.prepend() in step.js make the opening of a step fail. I will include a PR for this 😄.

Cheers!

alexdaube:
I was testing with release **2.0.0-beta-24 in IE** and some errors occurred. [Issue 238](https://github.com/shipshapecode/shepherd/issues/238)

Basically putting the `babel/env` back in the webpack config seem make IE load the lib instead of giving us the awful `Syntax Errors` message.

Also, [ParentNode.prepend](https://developer.mozilla.org/fr/docs/Web/API/ParentNode/append) is not fully supported by all the browsers, so it's probably better not to use it since it doesn't seem to be needed in this case.

Cheers!

Resolves #238

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -219,7 +219,7 @@ export class Step extends Evented {
       const title = document.createElement('h3');
       title.classList.add('shepherd-title');
       title.innerHTML = `${this.options.title}`;
-      header.prepend(title);
+      header.appendChild(title);
       element.classList.add('shepherd-has-title');
     }
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -175,12 +175,7 @@ module.exports.push({
         include: [
           path.resolve(__dirname, 'src/js')
         ],
-        use: [{
-          loader: 'babel-loader',
-          options: {
-            'presets': ['@babel/env']
-          }
-        }]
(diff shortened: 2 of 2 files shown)


rwwagner90:
Hi @alexdaube thanks for letting us know about this, and thanks for the PR! I think the issue was we didn't have a browserslist config, so it was not supporting IE as it should. I just added one. Can you see if that fixes things for you? I think we may not need the webpack config changes, now that we have the browserslistrc.

alexdaube:
On it @rwwagner90 👍.

alexdaube:
@rwwagner90 .  I guess I didn't look well enough earlier, as the babel-preset config was moved into the **babel.config.js**. The **browserslistrc** seem to be resolving the lib loading issue in IE, which is excellent. For `header.prepend(title);`, the issue is still present.

PR #244, opened 2018-09-11
Rename `Tour.options.defaults` to `Tour.options.defaultStepOptions`.

BrianSipple:
Closes [#240](https://github.com/shipshapecode/shepherd/issues/240).

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "rewrite-paths": "replace 'SF:.*src' 'SF:src' coverage/lcov.info",
     "start": "yarn watch",
     "start-test-server": "http-server",
+    "test": "yarn test:ci",
     "test:build": "cross-env NODE_ENV=test yarn build",
     "test:ci": "yarn test:unit:ci && yarn test:cy:ci",
     "test:cy:ci": "yarn test:build && start-server-and-test start-test-server http://localhost:8080 cy:run",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -34,11 +34,12 @@
     "rewrite-paths": "replace 'SF:.*src' 'SF:src' coverage/lcov.info",
     "start": "yarn watch",
     "start-test-server": "http-server",
+    "test": "cross-env NODE_ENV=test webpack --config webpack.test.config.js && yarn mocha-headless-chrome",
```
rwwagner90: I don't know if I agree with this change. I typically run `yarn test:ci` as the full test suite, to run all unit and acceptance tests, so if you want a `yarn test` command, we should just have `"test": "yarn test:ci"`, but I'm also not sure we need a `yarn test` command, since we can just use `yarn test:ci`. I'm okay with it though, if we make it run all the tests.

BrianSipple: I feel like it's useful to have a plain `test` command so new developers can instantly see which one is the default -- but I'm totally open to suggestions as to what it should say 😄!  

My original attempt was to just run `mocha-headless-chrome`, but that wasn't rebuilding the code as I made changes -- so that's how I ended up with the longer set of commands. I just updated again to use `yarn test:ci` instead.

PR #257, opened 2018-09-29
Change localhost port for cypress tests and document how its used

BrianSipple:
I still want to try and get acceptance tests working with live reloading, but in the meantime, I was able to make a few tweaks to the Cypress configuration and the way we're documenting it.


package.json
```diff
@@ -33,12 +33,12 @@
...
     "rewrite-paths": "replace 'SF:.*src' 'SF:src' coverage/lcov.info",
     "start": "yarn watch",
-    "start-test-server": "http-server",
+    "start-test-server": "http-server -p 9002",
     "test": "yarn test:ci",
+    "test:acceptance": "yarn test:build && start-server-and-test start-test-server http://localhost:9002 cy:open",
     "test:build": "cross-env NODE_ENV=test yarn build",
     "test:ci": "yarn test:unit:ci && yarn test:cy:ci",
-    "test:cy:ci": "yarn test:build && start-server-and-test start-test-server http://localhost:8080 cy:run",
-    "test:cy:watch": "start-server-and-test start-test-server http://localhost:8080 cy:open",
```
rwwagner90: Can we keep `test:cy:watch` please? I'd like to keep the format `test:{unit || cy}:{watch || ci}`

test/cypress/integration/test.acceptance.js
```diff
@@ -7,7 +7,7 @@ describe('Shepherd Acceptance Tests', () => {
   beforeEach(() => {
     Shepherd = null;
 
-    cy.visit('http://localhost:8080/test/dummy/', {
```
rwwagner90: I think we had to manually set this here, rather than using `baseURL` to make sure that things worked for both the demo app and the test app.

package.json
BrianSipple: Ah, I wasn't connecting `cy` with `unit` (the way one would relate `acceptance`). But that makes sense now 👌

test/cypress/integration/test.acceptance.js
BrianSipple: I'm not sure I'm following. Do you mean running the demo app with `yarn start`? It doesn't seem like that would interfere with `cy`

BrianSipple: Updated

package.json
BrianSipple: Updated

test/cypress/integration/test.acceptance.js
rwwagner90: I think it was to support the demo app, what we deploy for the demo.

rwwagner90: We can merge and see if the demo still looks good

PR #260, opened 2018-10-03
Support placement values and modifiers according to popper.js API (#259)

### LINKED ISSUE #259: Not all of popper.js placement values are supported
From the source code I see extracting _top|left|right|bottom|center_ values from the _attachTo_ option.
According to the **popper.js** API the _placement_ option may have the following values:
- auto
- top
- bottom
- left
and each value can have a variation: _-start_ or _-end_

Issues:
1. the _center_ value is not supported by **popper.js**
2. the _auto_ value is not supported by **shepherd.js**
3. none of the modifiers are supported by **shepherd.js**

tedbeer:
**popper.js** placement values are: _auto|top|bottom|left|right_ and modifiers are _-start|-end_

Closes #259

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -25,7 +25,7 @@ export function parsePosition(position) {
     return null;
   }
 
-  const positionRe = /^(.+) (top|left|right|bottom|center)$/;
+  const positionRe = /^(.+) ((auto|top|left|right|bottom)(-start|-end)?)$/;
   const matches = positionRe.exec(position);
 
   if (!matches) {


rwwagner90:
@BrianSipple will merging this mess up your tippy refactor stuff and/or does the tippy refactor render this change obsolete? We should probably just pass the directions in, without checking against the regex and let the libraries that consume those directions spit out errors, right?

BrianSipple:
@rwwagner90 There are a few pieces to that....

Tippy _does_ support a `placement` option directly. However, we're currently parsing the placement setting out of the `Step.attachTo` option and merging it in to the `tippy.popperOptions` hash. 

Ideally, we change `attachTo` to only be concerned with the element (see https://github.com/shipshapecode/shepherd/projects/2#card-13473896). But that's the kind of larger change we decided to hold off on until _after_ the initial Tippy integration. 

So with all that said... I think this change is good to go. (Thanks @tedbeer!)

PR #261, opened 2018-10-03
Tippy.js Integration

### LINKED ISSUE #255: Integrate Tippy for step modal creation
https://atomiks.github.io/tippyjs/

BrianSipple:
Closes #255 

Replaces our direct usage of Popper.js with [Tippy.js](https://atomiks.github.io/tippyjs/), which uses Popper under the hood and provides very robust animation support.

## Main changes

- Interface with Tippy instead of Popper
- Remove the `Step._attach` method. Tippy handles this automatically via its `appendTo` option.
    + This also removes the need for a `renderLocation` option.
- Remove usage of `Step.attachTo.modifiers`.
    + This should be left to `Step.popperOptions.modifiers`.
- Update styles to target Tippy's elements instead of Popper's.
- Update the demo app, documentation, and tests to show how to use Tippy. 

## Instructions for running
- Checkout the branch
- Run `yarn start`


docs/welcome/index.html
```diff
@@ -35,11 +35,11 @@ <h2>Guide your users through a tour of your app.</h2>
       <div class="hero-including">
         <h3>Including</h3>
 
-        <script type="text/plain" class="language-markup" id="hero-including-code">
+        <!-- <script type="text/plain" class="language-markup" id="hero-including-code">
```
rwwagner90: Why are we commenting out the usage docs here?

src/js/step.js
```diff
@@ -184,35 +183,20 @@ export class Step extends Evented {
...
+    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}'>`);
 
-    if (renderLocation) {
-      if (renderLocation instanceof HTMLElement) {
-        return renderLocation.appendChild(element);
-      }
-      if (isString(renderLocation)) {
-        return document.querySelector(renderLocation).appendChild(element);
-      }
+    if (this.options.idAttribute) {
```
rwwagner90: What is this `idAttribute` stuff?

src/js/step.js
```diff
@@ -299,13 +259,13 @@ export class Step extends Evented {
   destroy() {
     if (isElement(this.el) && this.el.parentNode) {
       this.el.parentNode.removeChild(this.el);
-      delete this.el;
+      this.el = null;
```
rwwagner90: Why set to `null` instead of deleting?

src/js/step.js
```diff
@@ -328,10 +288,9 @@ export class Step extends Evented {
       this.target.classList.remove('shepherd-enabled', 'shepherd-target');
     }
 
-    if (this.popper) {
-      this.popper.destroy();
+    if (this.tooltipElem) {
+      this.tooltipElem.hide();
```
rwwagner90: So tippy will destroy all its DOM when we cancel the tour without us manually destroying here?

src/js/step.js
```diff
@@ -413,7 +372,7 @@ export class Step extends Evented {
    *
    * @private
    */
-  _setUpButtons() {
+  _setupButtonOpts() {
```
rwwagner90: I've let some renaming slide, but this one seems arbitrary to me. Isn't a function that sets up buttons being named `setUpButtons` okay? 😛

src/js/step.js
```diff
@@ -450,14 +409,13 @@ export class Step extends Evented {
 
     document.body.setAttribute('data-shepherd-step', this.id);
 
-    this.setupPopper();
```
rwwagner90: Shouldn't we need to set up tippy on show still?

src/js/step.js
```diff
@@ -450,14 +409,13 @@ export class Step extends Evented {
...
-
     if (this.options.scrollTo) {
       setTimeout(() => {
         this.scrollTo();
       });
     }
 
+    this.tooltipElem.show();
```
rwwagner90: We should be setting up these tooltips when show is called, right? I think we need some setup code to replace `setupPopper` here. The idea is to create tooltips/poppers as needed, not to create them all at once, so they were being set up on show, and torn down when we cancel the tour, to minimize both the number of operations, and the amount of DOM we have lying around.

src/js/step.js
BrianSipple: Slightly above this line we do:

```js
    if (!this.el) {
      this.setupElements();
    }
```

`this.el` refers to the content of the tooltip. `this.setupElements` creates, that, and it also creates the tippy instance.

src/js/step.js
BrianSipple: See https://github.com/shipshapecode/shepherd/pull/261#discussion_r222499052. 

I'm fairly certain everything is happening as you described.

docs/welcome/index.html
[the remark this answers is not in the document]
rwwagner90: Fair enough 👍

src/js/step.js
BrianSipple: Totally. I don't remember why I did that, tbh. Perhaps I thought is was returning opts instead of setting `this.options.buttons` directly 🙃.

In any case, I changed it back.

src/js/step.js
BrianSipple: ... but then again, it looks like that's part of `Tour.done`, too. A few lines later we do:

```js
    if (Shepherd.activeTour) {
      Shepherd.activeTour.steps.forEach((step) => {
        step.destroy();
      });
    }
```

Running through the tour in the demo, all tippy elements and DOM changes do, indeed, get cleaned up.

docs/welcome/index.html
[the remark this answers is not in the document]
chuckcarpenter: okay, final just shouldn't have commented code.

docs/welcome/js/welcome.js
```diff
@@ -1,19 +1,22 @@
...
-  init = function() {
-    return setupShepherd();
-  };
+  function init() {
+    const shepherd = setupShepherd();
 
-  setupShepherd = function() {
-    var shepherd;
-    shepherd = new Shepherd.Tour({
+    setTimeout(() => {
```
chuckcarpenter: why?

src/js/step.js
```diff
@@ -102,15 +99,17 @@ export class Step extends Evented {
...
       'scrollTo',
+      'setupElements',
       'show'
     ]);
     this.setOptions(options);
     this.bindAdvance = bindAdvance.bind(this);
     this.bindButtonEvents = bindButtonEvents.bind(this);
     this.bindCancelLink = bindCancelLink.bind(this);
-    this.setupPopper = setupPopper.bind(this);
+    this.setupTooltipElem = setupTooltipElem.bind(this);
```
chuckcarpenter: Agree with steps that are agnostic to the lib as we can

src/js/step.js
chuckcarpenter: ID has been null in recent versions, does this fix that?

docs/welcome/js/welcome.js
```diff
@@ -25,7 +28,7 @@
           action: shepherd.next,
           text: 'Next'
         }
-      ]
+      ],
```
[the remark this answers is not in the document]
BrianSipple: 👍

docs/welcome/js/welcome.js
BrianSipple: This was to initiate a slight delay before starting the tour -- so the first entrance animation would be a bit more noticeable.

src/js/step.js
BrianSipple: `delete` has the effect of setting things to `undefined`, whereas setting to `null` is us saying that we aren't concerned about the reference anymore. 

I [tend to favor the latter](https://medium.com/@arieldi/anyway-stop-setting-your-variables-to-undefined-difference-undefined-vs-null-25f15d907fb4), generally -- but it also felt more consistent with the way we're cleaning up in other parts of our code.

src/js/step.js
rwwagner90: The idea is that `done` calls all that, yeah.

rwwagner90: Shepherd did not use to cleanup after itself, so I just want to make sure things are cleaned up. If they are, we're good.

rwwagner90: May just want to move things back to `done`. The less moving and changing the better 😄

src/js/step.js
rwwagner90: I think the idea is it should be completely gone, which `delete` gives us. Suppose it's no big deal, but would be good if we minimized changes.

src/js/step.js
rwwagner90: If we don't already have some tests that cover this, we should write some tests to check that the tippy/popper for each step is not created until the step is shown, and that all `.tippy` `.shepherd` etc elements are gone after the tour is completed or cancelled. With those tests in place and passing, I'm fine with whatever here.

src/js/step.js
BrianSipple: 👍

docs/welcome/index.html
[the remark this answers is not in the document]
BrianSipple: Updated

docs/welcome/js/welcome.js
```diff
@@ -1,19 +1,22 @@
...
+    const shepherd = new Shepherd.Tour({
       defaultStepOptions: {
         showCancelLink: true
-      }
+      },
     });
+
     shepherd.addStep('welcome', {
-      text: ['Shepherd is a javascript library for guiding users through your app. It uses <a href="https://popper.js.org/">Popper.js</a>, another open source library, to position all of its steps.', 'Popper makes sure your steps never end up off screen or cropped by an overflow. Try resizing your browser to see what we mean.'],
+      text: ['Shepherd is a JavaScript library for guiding users through your app. It uses <a href="https://popper.js.org/">Popper.js</a>, another open source library, to position all of its steps.', 'Popper makes sure your steps never end up off screen or cropped by an overflow. Try resizing your browser to see what we mean.'],
```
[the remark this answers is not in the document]
BrianSipple: Updated

src/js/step.js
BrianSipple: @rwwagner90 Just updated this in the most recent commit.

src/js/step.js
chuckcarpenter: From what I'm seeing now, no. Gets set in a Step, but I don't see where in options this is passed or has a fallback.

BrianSipple: Yeah, I was seeing ids of `undefined` on tooltips when `options.idAttribute` didn't exist. I'm not sure why the diff show it against removing `renderLocation`, but the original code was doing:

```js
    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id='${this.options.idAttribute}>`);
```

BrianSipple: Admittedly, it's a bit of a random change. I'm definitely in favor or using https://github.com/shipshapecode/shepherd/pull/262 instead.

src/js/step.js
BrianSipple: Just updated to revert back to `delete`.

docs/welcome/js/welcome.js
```diff
@@ -72,9 +121,18 @@
...
       ]
     });
-    return shepherd.start();
-  };
 
-  init();
+    return shepherd;
+  }
+
+  function ready() {
```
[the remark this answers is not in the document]
BrianSipple: In a roundabout way. I wanted to make sure that our demo took advantage of the fact that tooltip entry can now be animated, so I added a bit of logic to make sure the DOM was fully rendered before kicking off the tour.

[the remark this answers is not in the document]
rwwagner90: Gotcha 👍

src/js/utils.js
```diff
@@ -55,74 +57,144 @@ export function parseShorthand(obj, props) {
...
-  if (isUndefined(opts.element)) {
-    attachment = 'top';
-    _setupCenteredPopper(opts);
+/**
+ * Passes `options.attachTo` to `_parseAttachToOpts` to get the correct `attachTo` format
+ * @returns {({} & {element, on}) | ({})}
+ * `element` is a qualified HTML Element
+ * `on` is a string position value
+ */
+export function parseAttachTo() {
```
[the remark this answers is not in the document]
BrianSipple: Well... we actually have a few functions in `utils` that [rely on being bound](https://github.com/shipshapecode/shepherd/blob/master/src/js/step.js#L113) to the `Step` instance. Ultimately, I'd like to break out `utils` into to a few different files that are more focused and organized. 

I made _this_ change, however, as a very minimal step in that direction.

src/js/step.js
```diff
@@ -302,16 +257,16 @@ export class Step extends Evented {
       delete this.el;
     }
 
-    if (this.popper) {
-      this.popper.destroy();
+    if (this.tooltipElem) {
+      this.tooltipElem.destroy();
+      this.tooltipElem = null;
```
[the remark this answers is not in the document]
BrianSipple: I see where you're coming from, but I think that might actually be a reason _not_ to place it outside of the `if` statement: `this.tooltipElem` should never be undefined, false, 0, etc -- and if something ever failed because of that, we'd probably benefit by having it do so and getting to the root of the cause.

src/js/step.js
```diff
@@ -183,34 +182,14 @@ export class Step extends Evented {
...
    * Creates Shepherd element for step based on options
    *
    * @private
-   * @return {HTMLElement} The DOM element for the step
+   * @return {HTMLElement} The DOM element for the step tooltip
    */
-  _createElement() {
-    const content = document.createElement('div');
+  _createTooltipContent() {
+    const contentContainer = document.createElement('div');
```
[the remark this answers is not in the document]
BrianSipple: No disagreement here 😄

src/js/utils.js
```diff
@@ -55,74 +57,144 @@ export function parseShorthand(obj, props) {
...
+    this.tooltipElem.destroy();
   }
 
-  const opts = this.getAttachTo();
-  opts.modifiers = opts.modifiers || {};
-  let attachment = opts.on || 'right';
-  opts.positionFixed = false;
+  const attachToOpts = this.parseAttachTo();
+
+  this.tooltipElem = _makeTippyInstance.call(this, attachToOpts);
```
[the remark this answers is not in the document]
BrianSipple: Ah, yep. This was one of my earliest changes when I figured it probably shouldn't be `this.popper` 😄. 

What do you think about `this.tooltip`? I feel like there's a lot to be gained by keeping the name closer to the actual concept, as opposed to our library du jour.

src/js/step.js
BrianSipple: Right on, but part of my rationale here was to prevent us from potentially calling the `setupPopper` function twice [as part of what `_show()` does](https://github.com/shipshapecode/shepherd/blob/e3e95ec893dfe0e4de505308e482a37c50efd057/src/js/step.js#L453).

src/js/step.js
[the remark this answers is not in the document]
chuckcarpenter: If that's the case, then should this be a try/catch then?

src/js/tour.js
```diff
@@ -52,22 +57,23 @@ export class Tour extends Evented {
    * @return {Step} The newly added step
    */
   addStep(arg1, arg2) {
-    let name, step;
+    let name, step, stepOptions;
```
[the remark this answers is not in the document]
BrianSipple: Updated

src/js/utils.js
```diff
@@ -55,74 +57,144 @@ export function parseShorthand(obj, props) {
...
 /**
- * Determines options for Popper and initializes the Popper instance
+ * Determines options for the Tippy.js tooltip and initializes it.
  */
-export function setupPopper() {
-  if (isUndefined(Popper)) {
-    throw new Error('Using the attachment feature of Shepherd requires the Popper.js library');
+export function setupTooltipElem() {
+  if (isUndefined(tippy)) {
+    throw new Error(missingTippy);
```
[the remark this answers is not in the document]
BrianSipple: Good point. I interpreted the original `popper` logic as an intent to always check, but if not, I can remove this.

src/js/tour.js
```diff
@@ -2,6 +2,8 @@ import { isFunction, isNumber, isString, isUndefined } from 'lodash';
 import { Evented } from './evented.js';
 import { Step } from './step.js';
 import { bindMethods } from './bind.js';
+import tippy from 'tippy.js';
```
chuckcarpenter: adding it here, means we don't need it in the HTML `<script>`

src/scss/tippy/_base-tooltips.scss
```diff
@@ -0,0 +1,28 @@
...
+ *   <div class="tippy-tooltip">
+ *     <div class="tippy-content">
+ *       [[ Content Here ]]
+ *     </div>
+ *   </div>
+ * </div>
+ */
+
+.tippy-popper {
+  @include popper-theme-arrows($arrow-size: $arrow-size, $arrow-pointer-events: $arrow-pointer-events);
```
chuckcarpenter: looks like they go here @rwwagner90

src/js/step.js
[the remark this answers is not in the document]
BrianSipple: Updated

webpack.config.js
```diff
@@ -180,11 +180,11 @@ module.exports.push({
     ]
   },
   externals: {
-    'popper.js': {
-      root: 'Popper',
-      commonjs2: 'popper.js',
-      commonjs: 'popper.js',
-      amd: 'popper.js'
+    'tippy.js': {
```
chuckcarpenter: erg, oh yeah, forgot we use this to keep it external. That make sense? We'd maybe want to manage version anyway?

src/scss/_base.scss
```diff
@@ -1,21 +1,15 @@
 @import 'variables';
 @import 'mixins';
 @import 'helpers';
+@import 'tippy';
 
 @include box-sizing();
 
 .shepherd-element {
-  @include popper-theme-arrows($arrow-size: $arrow-size, $arrow-pointer-events: $arrow-pointer-events);
```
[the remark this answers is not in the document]
BrianSipple: I didn't think this change (moving the `@include popper-theme-arrows` invocation to `scss/tippy/_base-tooltips.scss`) was breaking any of the current behavior. 

Essentially, we're already applying `popper-theme-arrows` by default and making it unconfigurable. I think it'd be good to make an issue for addressing that in the future.

src/js/utils.js
```diff
@@ -55,74 +57,144 @@ export function parseShorthand(obj, props) {
...
 
 /**
- * Merge the global popperOptions, and the local opts
- * @param {String} attachment The direction for attachment
- * @param {Object} opts The local options
- * @return {Object} The merged popperOpts object
+ * Generates the hash of options that will be passed to `Tippy` instances
+ * target an element in the DOM.
+ *
+ * @param {Object} baseAttachToOptions The local `attachTo` options
```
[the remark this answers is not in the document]
BrianSipple: Yeah, I think I originally did this because I had part of the code modifying the hash. It doesn't look like I finished that way, though 😄 -- so good call.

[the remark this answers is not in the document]
BrianSipple: Updated

src/scss/tippy/_base-tooltips.scss
BrianSipple: Yep, I noted that [here].(https://github.com/shipshapecode/shepherd/pull/261#discussion_r223237023)

I think there's a future story we need to make for figuring out exactly how we're offering custom theming (which Tippy could be able to help with, too!)

webpack.config.js
BrianSipple: @chuckcarpenter @rwwagner90 Could be good to make a note about this in our [`2.x` planning](https://github.com/shipshapecode/shepherd/projects/2)

src/js/step.js
BrianSipple: I'm not sure I'm following... are you saying this can this be resolved? I think `hide` and `done` are both doing what they're supposed to now.

src/js/step.js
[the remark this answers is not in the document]
BrianSipple: I guess what I'm saying is we should be able to have absolute confidence and control over what `this.tooltipElem` -- so we shouldn't need to get _more_ advance with the checking either. 

If something fails in the future because someone has `this.tooltipElem` set to `0`, that's something we'd want to address then and there to discover how that could happen.

src/scss/tippy/_base-tooltips.scss
rwwagner90: 👍

src/scss/_base.scss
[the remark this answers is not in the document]
rwwagner90: It is configurable. All the variables passed to it can be overridden.

src/js/utils.js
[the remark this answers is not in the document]
rwwagner90: I just want to avoid having a bunch of files. I agree there could be potentially a better split, so this is fine for now.

src/js/utils.js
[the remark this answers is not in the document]
rwwagner90: I think `this.tooltip` would work, as long as we document that it is a `Tippy` instance 👍

src/js/step.js
rwwagner90: I'm not sure what I was saying here anymore 😂. We can resolve for now.

src/js/utils.js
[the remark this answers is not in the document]
BrianSipple: Cool -- just updated some of the commenting to clarify this.

src/scss/_base.scss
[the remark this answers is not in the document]
BrianSipple: Ah, true. But yeah, I think this can still be overridden by the user. See the way the [demo is doing exactly that](https://github.com/shipshapecode/shepherd/pull/261/files#diff-e1071944dd148ea7d2caef3b5fadfe44R124).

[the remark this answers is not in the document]
BrianSipple: The only difference is the need to target [Tippy's class/markup structure](https://atomiks.github.io/tippyjs/#creating-custom-themes) instead of `shepherd-element` directly.

I want to add more documentation around this and tweak some of the custom theming that Shepherd offers, but that feels like more of a [follow-up task](https://github.com/shipshapecode/shepherd/projects/2).

webpack.config.js
BrianSipple: I'd prefer to be more constrained and bundle a specific version of Tippy. That way, we know exactly what code we're supporting instead of creating an infinite possibility space 😄.

src/scss/_base.scss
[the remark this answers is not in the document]
rwwagner90: Follow up is fine, but we need to support theming via some easy method before releasing 2.0.

webpack.config.js
rwwagner90: @BrianSipple that sounds very logical and works for me. With that in mind, should we remove this from being external here?

BrianSipple: 🤔 I'd rather take it on as a follow-up task.

BrianSipple: https://github.com/shipshapecode/shepherd/issues/263

rwwagner90: Follow up is fine, as long as it's a quick follow up 😄

PR #262, opened 2018-10-05
use uniqueId for step ID prop over idAttribute

chuckcarpenter:
the ID for the main step DIV is always `undefined` currently, this sets it more explicitly as the options hash did not include `idAttribute`.

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -212,7 +212,7 @@ export class Step extends Evented {
   _createElement() {
     const content = document.createElement('div');
     const classes = this.options.classes || '';
-    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id="${this.options.idAttribute}"}>`);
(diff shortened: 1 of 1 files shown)


src/js/step.js
```diff
@@ -212,7 +212,7 @@ export class Step extends Evented {
   _createElement() {
     const content = document.createElement('div');
     const classes = this.options.classes || '';
-    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id="${this.options.idAttribute}"}>`);
+    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id="step-${this.options.id}-${uniqueId()}"}>`);
```
rwwagner90: Is `id` something we should still allow users to set? IIRC before we refactored `id` was the `id` passed to creating the step right? Or was that just `data-id`? If we do not want/need to let users set `id`, this seems fine to me 👍

chuckcarpenter: There's no reason to let the ID be set by users, since the problem I see with multiple tours are `data-id`s that are not unique. This makes sure that they will be.

PR #264, opened 2018-10-08
Update screenshot of intro step in README

BrianSipple:


BrianSipple:
Good catch. I could have sworn I pulled the latest from master before grabbing that. Anyway... updated.

PR #265, opened 2018-10-09
Implement `Tour.hide`

BrianSipple:
Closes [#249](https://github.com/shipshapecode/shepherd/issues/249).

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -292,10 +292,10 @@ export class Step extends Evented {
 
   /**
    * Check if the step is open and visible
-   * @return {*|boolean} True if the step is open and visible
+   * @return {boolean} True if the step is open and visible
    */
   isOpen() {
-    return this.el && !this.el.hidden;
+    return Boolean(this.el && !this.el.hidden);
   }
 
   /**
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -149,6 +149,17 @@ export class Tour extends Evented {
   }
 
   /**
+   * Hide the current step
+   */
+  hide() {
+    const currentStep = this.getCurrentStep();
+
+    if (currentStep) {
+      return currentStep.hide();
+    }
+  }
+
+  /**
    * Go to the next step in the tour
    * If we are at the end, call `complete`
    */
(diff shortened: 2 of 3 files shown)


src/js/step.js
```diff
@@ -295,7 +295,7 @@ export class Step extends Evented {
    * @return {*|boolean} True if the step is open and visible
    */
   isOpen() {
-    return this.el && !this.el.hidden;
+    return !!(this.el && !this.el.hidden);
```
chuckcarpenter: I've kind of feel `Boolean()` reads better. Why do I feel like @rwwagner90 has an opinion on this one...

src/js/utils/tooltip-defaults.js
```diff
@@ -1,7 +1,8 @@
 export const defaults = {
   trigger: 'manual',
   animation: 'fade',
-  delay: 200,
+  // delay: 200,
```
chuckcarpenter: remove

src/js/step.js
rwwagner90: I agree with @chuckcarpenter here, `Boolean` is much more readable than `!!`

rwwagner90: We should also change `@return {*|boolean}` to be `@return {boolean}` since we are ensuring it's always boolean now.

src/js/tour.js
```diff
@@ -148,6 +148,17 @@ export class Tour extends Evented {
     return this.currentStep;
   }
 
+  /**
+   * Hide the current step
+   */
+  hide() {
+    const currentStep = this.getCurrentStep();
+
+    if (currentStep && typeof currentStep.hide === 'function') {
```
rwwagner90: We should use `isFunction` from lodash. We're using a few lodash functions throughout the code. Webpack tree shakes things for us, so if you just import the function you need our build stays small. Also, is there a case when then step would not have a `hide` function on it? If not, we could probably remove this check.

rwwagner90: @BrianSipple if you can update or remove this, we'll be good to go 👍

BrianSipple: Agreed that the function check probably shouldn't be necessary. Updated by removing it.

PR #279, opened 2018-10-17
Allow exit animations to play before step tooltip disappears.

BrianSipple:
Closes https://github.com/shipshapecode/shepherd/issues/277

This removes our own manual management of `this.el`'s appearance/presence
in the DOM -- since tippy is handling it already. 

This allows the tooltip's exit animation to play, as opposed to having the element disappear instantly before we even call `tippy.hide`.

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -247,14 +247,10 @@ export class Step extends Evented {
    * Triggers `destroy` event
    */
   destroy() {
-    if (isElement(this.el) && this.el.parentNode) {
-      this.el.parentNode.removeChild(this.el);
-      delete this.el;
-    }
-
     if (this.tooltip) {
       this.tooltip.destroy();
       this.tooltip = null;
+      this.el = null;
     }
 
     this.trigger('destroy');
@@ -265,12 +261,6 @@ export class Step extends Evented {
    */
   hide() {
     this.trigger('before-hide');
-
-    if (this.el) {
-      this.el.hidden = true;
-      // We need to manually set styles for < IE11 support
-      this.el.style.display = 'none';
-    }
 
     document.body.removeAttribute('data-shepherd-step');
 
@@ -290,7 +280,11 @@ export class Step extends Evented {
    * @return {boolean} True if the step is open and visible
    */
   isOpen() {
-    return Boolean(this.el && !this.el.hidden);
+    return Boolean(
+      this.tooltip &&
+      this.tooltip.state &&
+      this.tooltip.state.isVisible
+    );
   }
 
   /**
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -250,7 +250,15 @@ export class Step extends Evented {
     if (this.tooltip) {
       this.tooltip.destroy();
       this.tooltip = null;
+    }
+
+    if (this.el && this.el instanceof HTMLElement) {
+      this.el.parentNode.removeChild(this.el);
       this.el = null;
+    }
+
+    if (this.target) {
+      this.target.classList.remove('shepherd-enabled', 'shepherd-target');
     }
 
     this.trigger('destroy');
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -252,7 +252,7 @@ export class Step extends Evented {
       this.tooltip = null;
     }
 
-    if (this.el && this.el instanceof HTMLElement) {
+    if (isElement(this.el) && this.el.parentNode) {
       this.el.parentNode.removeChild(this.el);
       this.el = null;
     }
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -1,4 +1,4 @@
-import { isFunction, isNumber, isString, isUndefined } from 'lodash';
+import { isFunction, isNumber, isString, isUndefined, isEmpty } from 'lodash';
 import { Evented } from './evented.js';
 import { Step } from './step.js';
 import { bindMethods } from './bind.js';
@@ -112,17 +112,11 @@ export class Tour extends Evented {
    * @param {String} event The event name to trigger
    */
   done(event) {
-    if (this.currentStep) {
-      this.currentStep.hide();
+    if (!isEmpty(this.steps)) {
+      this.steps.forEach((step) => step.destroy());
     }
 
     this.trigger(event);
-
-    if (Shepherd.activeTour) {
-      Shepherd.activeTour.steps.forEach((step) => {
-        step.destroy();
-      });
-    }
 
     Shepherd.activeTour = null;
     document.body.classList.remove('shepherd-active');
@@ -224,6 +218,10 @@ export class Tour extends Evented {
    * @param {Boolean} forward True if we are going forward, false if backward
    */
   show(key = 0, forward = true) {
+    if (this.currentStep) {
+      this.currentStep.hide();
+    }
+
     this._setupActiveTour();
 
     const step = isString(key) ? this.getById(key) : this.steps[key];
@@ -258,17 +256,12 @@ export class Tour extends Evented {
   }
 
   /**
-   * If we have a currentStep, the tour is active, so just hide the step and remain active.
-   * Otherwise, make the tour active.
+   * Make this tour "active"
    * @private
    */
   _setupActiveTour() {
-    if (this.currentStep) {
-      this.currentStep.hide();
-    } else {
-      document.body.classList.add('shepherd-active');
-      this.trigger('active', { tour: this });
-    }
+    document.body.classList.add('shepherd-active');
+    this.trigger('active', { tour: this });
 
     Shepherd.activeTour = this;
   }
--- a/test/cypress/integration/test.acceptance.js
+++ b/test/cypress/integration/test.acceptance.js
@@ -224,20 +224,5 @@ describe('Shepherd Acceptance Tests', () => {
         cy.get('.shepherd-step-element').should('have.length', 1);
       });
     });
-
-    describe('Cleaning up', () => {
-      let tour;
-
-      beforeEach(() => {
-        tour = setupTour(Shepherd);
-        tour.start();
-      });
-
-      it('renders no steps after the tour completes', () => {
-        tour.complete();
-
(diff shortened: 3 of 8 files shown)


test/unit/test.step.js
```diff
@@ -95,111 +106,108 @@ describe('Step', () => {
   });
 
   describe('bindAdvance()', () => {
```
BrianSipple: These changes are mainly due to the fact that `Step.isOpen` is no longer reading `this.el.hidden` to determine that.

... but I also took the opportunity to abstract some of the duplication to a `beforeEach` 😄.

src/js/step.js
```diff
@@ -247,16 +247,20 @@ export class Step extends Evented {
...
-      delete this.el;
-    }
-
     if (this.tooltip) {
       this.tooltip.destroy();
       this.tooltip = null;
     }
 
+    if (this.el && this.el instanceof HTMLElement) {
+      this.el.remove();
```
rwwagner90: We need to set this back to the old code. `remove` does not work in IE.

src/js/step.js
```diff
@@ -266,16 +270,10 @@ export class Step extends Evented {
...
-      this.el.hidden = true;
-      // We need to manually set styles for < IE11 support
-      this.el.style.display = 'none';
-    }
-
     document.body.removeAttribute('data-shepherd-step');
 
     if (this.target) {
-      this.target.classList.remove('shepherd-enabled', 'shepherd-target');
+      this._removeClassesFromTarget();
```
rwwagner90: Can we please just leave this as `classList.remove`? Moving one line into a function is actually adding more code, not to mention it's harder to read because now I have to search for this function to see what it is doing.
```suggestion
      this.target.classList.remove('shepherd-enabled', 'shepherd-target');
```

src/js/tour.js
```diff
@@ -258,17 +262,12 @@ export class Tour extends Evented {
...
 
   /**
-   * If we have a currentStep, the tour is active, so just hide the step and remain active.
-   * Otherwise, make the tour active.
+   * Make this tour "active"
    * @private
    */
   _setupActiveTour() {
-    if (this.currentStep) {
-      this.currentStep.hide();
```
rwwagner90: I'm a little unclear on why we had this hide in the first place, but can you please elaborate on why it is no longer needed?

test/unit/test.step.js
```diff
@@ -7,24 +7,27 @@ const { assert } = chai;
 import Shepherd from '../../src/js/shepherd.js';
 import { Step } from '../../src/js/step.js';
 import { Tour } from '../../src/js/tour.js';
+import tippy from 'tippy.js';
```
rwwagner90: Shouldn't we be importing Tippy in one of the other files above we are already importing?

test/unit/test.tour.js
```diff
@@ -7,20 +7,25 @@ import tippy from 'tippy.js';
...
     scrollTo: true
   };
 
+  beforeEach(() => {
+    tippy.disableAnimations();
+  });
+
   afterEach(() => {
-    instance.cancel();
+    instance.complete();
```
rwwagner90: Is there a reason we need this to be `complete` and not `cancel`? I think both should call `done` right?

src/js/step.js
```diff
@@ -290,7 +288,11 @@ export class Step extends Evented {
    * @return {boolean} True if the step is open and visible
    */
   isOpen() {
-    return Boolean(this.el && !this.el.hidden);
+    return Boolean(
+      this.tooltip &&
```
rwwagner90: Just to confirm, we're just using built in Tippy stuff here now, instead of `hidden` right?

src/js/tour.js
```diff
@@ -113,7 +113,7 @@ export class Tour extends Evented {
    */
   done(event) {
     if (this.currentStep) {
-      this.currentStep.hide();
+      this.currentStep.destroy();
```
rwwagner90: What about destroying all the rest of the steps? We want to make sure we clean up all the event listeners and such Tippy is adding.

src/js/step.js
BrianSipple: True... but the reason I moved this to a function is because I'm calling it from two separate locations now.

src/js/tour.js
BrianSipple: Exactly -- it doesn't seem like it would be a part of `_setupActiveTour`. 

`_setupActiveTour`, however, is called as a part of `next.` So I just moved the `hide` call into `next` to make our intents more clear.

test/unit/test.step.js
BrianSipple: I'm not sure I'm following. Other files might be importing it, but I don't think that would make it global if that's what you mean?

src/js/step.js
BrianSipple: 👍

src/js/tour.js
BrianSipple: That's already being done in a few lines later: 

```js
    if (Shepherd.activeTour) {
      Shepherd.activeTour.steps.forEach((step) => {
        step.destroy();
      });
    }
```

I wondered why we wouldn't just do `this.steps.forEach(step => step.destroy());`... but it didn't seem to be a blocker here.

test/unit/test.tour.js
BrianSipple: They do, but cancel has a lot of functionality around displaying promps if the user has that set as an option. So `complete` felt better both semantically  _and_ to reduce the risk of unforseen side-effects.

src/js/step.js
BrianSipple: Updated

src/js/step.js
rwwagner90: Yeah, but it doesn't save any lines of code, it adds lines, since it was just one line. I prefer to keep it as it was.

test/unit/test.tour.js
rwwagner90: We should be making sure to not be setting `confirmCancel` to true, and this should be a non-issue.

src/js/tour.js
rwwagner90: It feels a bit odd to me to destroy one step. I think the `forEach` destroying all should be all we need right? I could be wrong, but if that is the case, feel free to nix this 👍

test/unit/test.step.js
rwwagner90: I think I was thinking perhaps it was added onto `Shepherd` for instance, to access via `Shepherd.tippy`. If it's not available elsewhere, this is fine 👍

test/unit/test.tour.js
BrianSipple: Oh, no, it's not an issue now. I'm just thinking it reduces risk of coupling unexpected behavior in the future.

src/js/step.js
BrianSipple: Updated

src/js/tour.js
BrianSipple: No, I think you're right. I wasn't sure if there was a specific reason for having the two operations split out like they are, so I didn't want to change it -- but I'll give the single `forEach` a shot now.

test/unit/test.tour.js
rwwagner90: Sure, just feels a bit arbitrary to me, but this is fine 👍

src/js/step.js
```diff
@@ -247,16 +247,20 @@ export class Step extends Evented {
...
-      this.el.parentNode.removeChild(this.el);
-      delete this.el;
-    }
-
     if (this.tooltip) {
       this.tooltip.destroy();
       this.tooltip = null;
     }
 
+    if (this.el && this.el instanceof HTMLElement) {
```
rwwagner90: This should be the same as it was before.
```suggestion
   if (isElement(this.el) && this.el.parentNode) {
```

src/js/tour.js
BrianSipple: Updated

src/js/tour.js
```diff
@@ -112,18 +112,12 @@ export class Tour extends Evented {
...
   done(event) {
-    if (this.currentStep) {
-      this.currentStep.hide();
+    if (!isEmpty(this.steps)) {
+      this.steps.forEach((step) => step.destroy());
     }
 
     this.trigger(event);
 
-    if (Shepherd.activeTour) {
```
rwwagner90: Isn't this removed code here effectively doing the same thing as the added code above? I'm curious if there is a benefit to switching to `this.steps.forEach` over `Shepherd.activeTour.steps.forEach`?

test/cypress/integration/test.acceptance.js
```diff
@@ -224,16 +224,5 @@ describe('Shepherd Acceptance Tests', () => {
         cy.get('.shepherd-step-element').should('have.length', 1);
       });
     });
-
-    describe('Cleaning up', () => {
```
rwwagner90: Shouldn't this test still work?

src/js/tour.js
BrianSipple: My answer to the second question is that my answer to the first question is "I don't know" 😄 

If it's not necessary to go through `Shepherd.activeTour.steps`, I think using `this.steps` is a lot more explicit about our intent.

test/cypress/integration/test.acceptance.js
BrianSipple: https://github.com/shipshapecode/shepherd/pull/279#pullrequestreview-165925406

src/js/tour.js
rwwagner90: I think the idea is you would only want to destroy steps from the currently active tour. `this.steps` accomplishes roughly the same thing, but I think the main difference is if there is no active tour when `done` is called, there is no need to bother tearing anything down.

BrianSipple: 🤔 But if `done` is called on the tour instance, shouldn't that instance only care about itself?

It seems like this would risk a) not tearing down the steps of the instance that a user calls `done` on, and b) accidentally tearing down steps of a separate instance.

Both of those seem like things we -- or users -- wouldn't want.

PR #282, opened 2018-10-21
Change `data-id` to `data-shepherd-step-id` on Step content elements.

BrianSipple:
clashResiliance++;

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -190,7 +190,7 @@ export class Step extends Evented {
   _createTooltipContent() {
     const content = document.createElement('div');
     const classes = this.options.classes || '';
-    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id="step-${this.options.id}-${uniqueId()}"}>`);
+    const element = createFromHTML(`<div class='${classes}' data-shepherd-step-id='${this.id}' id="step-${this.options.id}-${uniqueId()}"}>`);
     const header = document.createElement('header');
 
     if (this.options.title) {
--- a/test/unit/test.step.js
+++ b/test/unit/test.step.js
@@ -100,7 +100,7 @@ describe('Tour | Step', () => {
       it('shows step evoking method, regardless of order', () => {
         showTestStep.show();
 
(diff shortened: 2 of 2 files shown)


src/js/step.js
```diff
@@ -190,7 +190,7 @@ export class Step extends Evented {
   _createTooltipContent() {
     const content = document.createElement('div');
     const classes = this.options.classes || '';
-    const element = createFromHTML(`<div class='${classes}' data-id='${this.id}' id="step-${this.options.id}-${uniqueId()}"}>`);
+    const element = createFromHTML(`<div class='${classes}' data-shepherd-step-id='${this.id}' id="step-${this.options.id}-${uniqueId()}"}>`);
```
chuckcarpenter: I'm good w/ removing the `ID` attr, but since with multiple tours we can have shared ids, can we do something to make sure this value is unique?

BrianSipple: Then again, do we need uniqueness here? 

We were discussing having a `data-shepherd-active-tour` attribute on the body. If we implement that, someone could always use it as the context for querying unique steps -- even with shared ids.

BrianSipple: @chuckcarpenter @rwwagner90 To summarize, I'm thinking we merge this PR as-is, then open up another issue for adding `data-shepherd-active-tour` attribute and removing the extra `id` attribute on steps.

BrianSipple: @chuckcarpenter Where are you seeing that? I'm thinking we still need to implement it. Right now, it looks like we're just indicating the current step:

screen shot 2018-10-22 at 4 01 26 pm

chuckcarpenter: you're right. It's in the BM frontend right now. 

```
// TODO: add this in Shepherd.js
document.body.setAttribute('data-shepherd-active-tour', tourName);
```

chuckcarpenter: that said, should the tour id be part of the ID for the step?

BrianSipple: @chuckcarpenter [I don't think it would be necessary](https://github.com/shipshapecode/shepherd/pull/282#discussion_r226879920).

PR #292, opened 2018-11-05
Changes to support selecting elements in shadow DOM and configuring an overlay backdrop.

ericpias:
I really liked shepherd for my tour needs on my app.  However, we use web components in our web app (using Polymer 2/3 but that is not significant).  So I needed to be able to attach tour steps to elements found within the shadow dom of various components on the page.  

Two major changes: 
1. Make it work with web components and shadow dom.  Need to be able to attach (virtually) to sub-elements found within (possibly) nested shadow dom. 
2. Add support to optionally configure the tour to show an overlay backdrop to the tour steps. The overlay can be styled any way we want.

For change 1, the attachTo option syntax was extended a bit.  The change only supports selecting shadow dom elements using the object notation.  The element sub-property is extended to support an array of selectors.  Each selector selects into a DOM starting at 'document'.  Succeeding selectors select into the (open) shadowRoot of the found element. 

Change 2 just optionally creates absolutely positioned overlays for tour steps.  If there is no attachTo, then only a single overlay is created, otherwise 4 are created that border the target attach to element (top, right, bottom, left).  Classes and styles may be optionally configured as well as cancel tour on click.  This is documented in step.js.

My apologies for combining these changes together, but I really needed both.  Hopefully I can contribute this back to the project.

I have changed the demo page and acceptance tests.  I was not sure how to change the documentation, but am willing to do so if it helps.


rwwagner90:
Hi @ericpias thanks for the PR! I don't know much about shadow DOM. Does our existing overlay functionality not work for it?

ericpias:
Robert, 

The overlay feature that I added is somewhat independent of shadow dom.  However, when the attachTo element is in shadow dom, I create a virtual target element that exactly covers and tracks the attachTo element's position.  Tippy needs that virtual target element in order to work properly.  In this case, the virtual target element is used to absolutely position the backdrop overlay divs in the outer dom.  Otherwise, the real attachTo target is used.

All that being said, when you refer to the existing overlay functionality, can you either describe that or point me toward it in the doc.  The only thing I saw was in the demo where you were setting the opacity on the page (or something like that).  Is there something else that maybe I missed?

Eric

__________________________________________________________
Eric Pias | frevvo Inc. | 500 E. Main St, Branford, CT 06405 | http://www.frevvo.com <http://www.frevvo.com/>
m: 860.754.8480 | eric.pias@frevvo.com | Skype: ericpias

rwwagner90:
@ericpias apologies, apparently we have not implemented overlay functionality here, it is in [ember-shepherd](https://github.com/shipshapecode/ember-shepherd). @BrianSipple can we either move that functionality over here or perhaps look at using the functionality @ericpias added here? Whichever you guys think works better is fine with me 😃

rwwagner90:
@ericpias I just merged in the modal code, finally. Apologies for the delay. We're using svg masks for the modal. Can you tweak your PR here to work with that modal implementation or if you believe your solution is better, articulate what makes it better please?

rwwagner90:
@ericpias would you have some time to revisit this, with the modal implementation we added?

rwwagner90:
@ericpias we would love to support your use case, so please open a new PR and let us know how we can help!

PR #301, opened 2018-12-14
Integrate modal functionality that was originally in Ember Shepherd

BrianSipple:
There's still a bit that needs to be added to the test coverage here, but overall, I think covers what we want to move over, @rwwagner90


rwwagner90:
@BrianSipple looking good so far! We'll definitely want to port tests over as well before merging, but happy to see this functionality materializing here 😄

PR #302, opened 2018-12-19
Fix: stop ontouchstart events on overlay

chuckcarpenter:
This branch stops the ability on iOS (where overflow: hidden doesn't work) to pull the overlay window out of place and move the mask highlight out of place.

🚈

### DIFF
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -19,8 +19,6 @@ import {
 import {
   getModalMaskOpening,
   createModalOverlay,
-  preventModalBodyTouch,
-  preventModalOverlayTouch,
   positionModalOpening,
   closeModalOpening,
   classNames as modalClassNames,
@@ -160,8 +158,6 @@ export class Tour extends Evented {
     cleanupStepEventListeners.call(this);
     cleanupSteps(this.tourObject);
     cleanupModal.call(this);
-
-    window.removeEventListener('touchmove', preventModalBodyTouch, false);
 
     this.trigger(event);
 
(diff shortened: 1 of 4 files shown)


src/js/tour.js
```diff
@@ -159,6 +161,8 @@ export class Tour extends Evented {
     cleanupSteps(this.tourObject);
     cleanupModal.call(this);
 
+    window.removeEventListener('touchmove', preventModalBodyTouch, false);
```
BrianSipple: Could this be moved into [`cleanupStepEventListeners`](https://github.com/shipshapecode/shepherd/pull/302/files#diff-78c335d6579502a5ab234e509a14fba0R160)?

src/js/tour.js
```diff
@@ -328,6 +332,13 @@ export class Tour extends Evented {
       this._hideModalOverlay();
 
       document.body.appendChild(this._modalOverlayElem);
+
+      const overlay = document.querySelector('#shepherdModalOverlayContainer');
+      // Prevents window from moving on touch.
+      window.addEventListener('touchmove', preventModalBodyTouch, false);
+
+      // Allows content to move on touch.
+      overlay.addEventListener('touchmove', preventModalOverlayTouch, false);
```
BrianSipple: Similar question: Perhaps we could move this to the `addStepEventListeners`  util.

BrianSipple:
Some questions regarding where we call this from, but overall LGTM 👍

chuckcarpenter:
@BrianSipple that makes sense to me and I've updated.

PR #305, opened 2019-01-03
fix: remove class when modal hidden so elements are clickable

chuckcarpenter:
this branch adds a call to the `._hideModalOverlay()` when the main tour hide method is called so that the `.shepherd-modal-is-visible` class is removed and pointer events are restored.

### DIFF
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -192,6 +192,7 @@ export class Tour extends Evented {
     const currentStep = this.getCurrentStep();
 
     if (currentStep) {
+      this._hideModalOverlay();
       return currentStep.hide();
     }
   }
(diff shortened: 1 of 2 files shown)


src/js/tour.js
```diff
@@ -408,7 +409,9 @@ export class Tour extends Evented {
    * @private
    */
   _hideModalOverlay() {
-    document.body.classList.remove(modalClassNames.isVisible);
+    if (document.body.classList.contains(modalClassNames.isVisible)) {
+      document.body.classList.remove(modalClassNames.isVisible);
```
BrianSipple: Do we need the `if` check? Calling remove shouldn't fail if there's no match

chuckcarpenter: true, was just overdoing it I suppose

PR #308, opened 2019-01-06
Move modal to its own class

rwwagner90:
I still need to write some tests, now that this is its own class, but wanted to get some eyes on it beforehand.


BrianSipple:
👍

PR #312, opened 2019-01-13
setAttributeNS -> setAttribute, add modal utils tests

rwwagner90:


.codeclimate.yml
```diff
@@ -1,5 +1,11 @@
 version: "2"
 checks:
+  argument-count:
+    config:
+      threshold: 10
```
BrianSipple: After 5, I tend to think an object hash is in order. Totally don't want to debate this too heavily, though 😄.

src/js/utils/modal.js
```diff
@@ -113,23 +104,24 @@ function createModalOverlay() {
...
 function positionModalOpening(targetElement, openingElement) {
   if (targetElement.getBoundingClientRect && openingElement instanceof SVGElement) {
     const { x, y, width, height } = targetElement.getBoundingClientRect();
 
-    openingElement.setAttributeNS(null, 'x', x);
-    openingElement.setAttributeNS(null, 'y', y);
-    openingElement.setAttributeNS(null, 'width', width);
-    openingElement.setAttributeNS(null, 'height', height);
+    console.log('x', x);
+    console.log('y', y);
```
BrianSipple: Do we still want to be logging here?

src/js/utils/modal.js
```diff
@@ -159,6 +151,22 @@ function toggleShepherdModalClass(currentElement) {
...
+ * @param {string} height A string representing the height of the element
+ * @private
+ */
+function _setElementDimensions(element, x, y, width, height) {
+  element.setAttribute('x', x);
+  element.setAttribute('y', y);
+  element.setAttribute('width', width);
+  element.setAttribute('height', height);
+}
+
```
BrianSipple: To make this read a bit more explicitly when it's called, we could take an object hash. 

The function could then destructure the properties and it would read pretty much the same here too.

src/js/utils/modal.js
rwwagner90: Nope, good catch!

src/js/utils/modal.js
rwwagner90: Went with an in between option. Using a hash, but not destructuring, and instead using `Object.keys` that way, we can set any attributes we want.

.codeclimate.yml
rwwagner90: Yeah, I was just tired of Code Climate being so picky, but tend to agree 👍

PR #330, opened 2019-02-08
svg mask for the opening in browsers do not support getBoundingClientRect().x|y

yaxinr:
[Yandex browser](https://en.wikipedia.org/wiki/Yandex_Browser) not support getBoundingClientRect().x.


rwwagner90:
@yaxinr We support all major browsers. I've never heard of Yandex. According to the link you posted, it is supposedly chromium based, so it should work.

PR #357, opened 2019-04-18
[BugFix] Issue #347

jayjfletcher:
The _modalOverlayElem is destroyed upon cleanup, but is never rebuilt when restarting a tour.

Moved the creation of the modal from the constructor to the startup process of the tour. This way the modal is rebuilt for each run through the tour.

This is a fix for Issue #347

### DIFF
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -80,8 +80,6 @@ export class Tour extends Evented {
       })(event);
     });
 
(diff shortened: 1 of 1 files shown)


jayjfletcher:
Sorry I referenced the wrong issue # in my initial commit... 

This is for issue #347

PR #360, opened 2019-04-24
Add TypeScript definitions

superheri:
Trying to implement #359.

For now, only Step types are present as of a base. A couple of things would need to be clarified. 

Many TODOs are in the code @rwwagner90 if you have some time, could you give me your thoughts on this start? There are some things I'm not sure about, if you could check the TODOs that would be nice.

If you like how it goes, I would continue with the other files.

### DIFF
--- a/src/types/evented.d.ts
+++ b/src/types/evented.d.ts
@@ -1,9 +1,14 @@
-declare namespace Evented {
-    
-}
+import Tour from "./tour";
+import Step from "./step";
 
 declare class Evented {
-    //TODO superheri
+    on(event: string, handler: Function, ctx: Step | Tour): void;
+
+    once(event: string, handler: Function, ctx: Step | Tour): void;
+
+    off(event: string, handler: Function): boolean | void;
+
+    trigger(event: string): void;
 }
 
 export = Evented;
--- a/src/types/evented.d.ts
+++ b/src/types/evented.d.ts
@@ -1,0 +1,9 @@
+declare namespace Evented {
+    
+}
+
+declare class Evented {
+    //TODO superheri
+}
+
+export = Evented;
--- a/src/types/step.d.ts
+++ b/src/types/step.d.ts
@@ -13,7 +13,7 @@ declare namespace Step {
          * If you use the object syntax, element can also be a DOM element. If you don’t specify an attachTo the
          * element will appear in the middle of the screen.
          */
-        attachTo: StepOptionsAttachTo | string;
+        attachTo?: StepOptionsAttachTo | string;
 
         /**
          * An action on the page which should advance shepherd to the next step.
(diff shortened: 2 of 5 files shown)


rwwagner90:
@superheri This looks like a good start to me! I will take a more in depth look later this week.

rwwagner90:
@chuckcarpenter since you are likely more knowledgable on TS, perhaps you could take a look?

src/types/evented.d.ts
```diff
@@ -0,0 +1,9 @@
+declare namespace Evented {
+    
+}
+
+declare class Evented {
```
chuckcarpenter: I don't think you have to declare both for a class?

src/types/step.d.ts
```diff
@@ -0,0 +1,241 @@
...
+    interface StepOptionsWhen {
+        [key: string]: (() => void);
+    }
+}
+
+/**
+ * Class representing steps to be added to a tour
+ * @extends {Evented}
+ */
+declare class Step extends Evented {
```
chuckcarpenter: again, not clear why these would be split? https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html#classes

superheri: The reason I split the class we want to export and the sub types is that what we want to export is mostly the Step class. The other types are used to add types for the different properties of the main exported class. I based this on the explanations I received from a guy from Microsoft in the DefinitelyTyped official repo when I created type definitions for an other package : https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34714

superheri: Here is an example from TypeScript : https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html

rwwagner90:
@chuckcarpenter @superheri any updates on this?

superheri:
I was waiting for an answer before continuing. But if you want I can continue slowly during my free times.

rwwagner90:
@superheri an answer about what? @chuckcarpenter can you please weigh in?

superheri:
@rwwagner90 About the questions that @chuckcarpenter asked me directly about the code and the reason to do it that way.

chuckcarpenter:
@superheri I resolved my initial question based on your answer. If you'd like to start by adding the definitions, we can look at rewriting in TS over time. Sorry for the confusion!

chuckcarpenter:
@superheri I resolved my initial question based on your answer. If you'd like to keep working on this, we can look at rewriting in TS over time. Sorry for the confusion!

superheri:
Ok then, I can't promise you that I'll be very fast, but I'll continue to work on this.

superheri:
@rwwagner90 Seems like it for now. From what I see not everything needs typings since not everything is exposed. Maybe there are some things that should be changed, feel free to tell me if there are anything you want me to change.

rwwagner90:
@chuckcarpenter do you mind reviewing again please?

rwwagner90:
@superheri this looks good to me! Have you tried this out with your project to ensure the typings behave as you expect? If so, I think we should be good to merge 😃

superheri:
@rwwagner90 I stopped working at my job where I this package is used so I no more have a real usage of it to test it out correctly. So I haven't tried it in a real case. I tried calling the different objects exposed individually in a really small project, but I don't think it is enough. Maybe if one of you has a more complicated demo, I could try that out with it and maybe you could try it out too so we will be sure it works fine before releasing it.

superheri:
@rwwagner90 Just tried it out with the welcome.js file whan converting it into TypeScript (Testing with TypeScript 3.5.1). (Just realised I had a demo to test it...) I adapted some things and it seems to work with that demo now. However, the demo does not represent every cases and maybe there are some that are not covered. 

But if you want we could try to deploy it as is for a first version and see if anyone has issues with this.

rwwagner90:
I think deploying as a first version sounds good 👍

PR #396, opened 2019-06-10
Remove addStepEventListeners call

### LINKED ISSUE #393: "TypeError: Property 'handleEvent' is not callable." in Firefox
In firefox on load and on scroll I receive this error in console:
> TypeError: Property 'handleEvent' is not callable.

I have debugged a little and found the reason.
util/dom.js
addStepEventListeners() lines 82, 83
https://github.com/shipshapecode/shepherd/blob/master/src/js/utils/dom.js#L82

this._onScreenChange is null. You can't register null handler.

rwwagner90:
I am not 100% sure if removing this is okay or not. @chuckcarpenter do you know why we had this call here? Is `addStepEventListeners` only necessary when using modal mode?

Fixes #393

### DIFF
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -12,7 +12,6 @@ import {
 } from './utils/cleanup';
 
 import {
-  addStepEventListeners,
   getElementForStep
 } from './utils/dom';
 
@@ -302,7 +301,6 @@ export class Tour extends Evented {
 
     this.currentStep = null;
     this._setupActiveTour();
-    addStepEventListeners.call(this);
     this.next();
   }
 


chuckcarpenter:
Looks like this is just a port of the Ember add-on, so not sure otherwise: https://github.com/shipshapecode/shepherd/commit/f1c4eeb05b84fa3f2431857391a8a45400b747f3

rwwagner90:
@chuckcarpenter `addStepEventListeners` is still called here https://github.com/shipshapecode/shepherd/blob/bef53420fc96a4bd08dee046afcfc5023244e34b/src/js/modal.js#L105

I think this extra call was maybe to ensure some of the touchmove stuff was setup immediately or something? If you can't see anything in `addStepEventListeners` that we *need* to run in the tour constructor, I say we remove it.

chuckcarpenter:
From my local testing in iOS and the docs site, this seems to work.

PR #401, opened 2019-06-12
Allow to specify scrollIntoView behavior #398

hheexx:
https://github.com/shipshapecode/shepherd/issues/398


src/js/step.js
```diff
@@ -4,6 +4,7 @@ import { Evented } from './evented.js';
 import 'element-matches';
 import { bindAdvance, bindButtonEvents, bindCancelLink, bindMethods } from './utils/bind.js';
 import { createFromHTML, setupTooltip, parseAttachTo } from './utils/general.js';
+import { isBoolean } from 'util';
```
rwwagner90: I don't believe `util` is a file and I also don't think we have an `isBoolean`.

PR #408, opened 2019-06-21
Add focus trap, to disallow tabbing outside the modal

rwwagner90:
This keyboard trap implementation was taken from https://bitsofco.de/accessible-modal-dialog/ and is related to #198


rwwagner90:
@chuckcarpenter do you think we should trap the keyboard for both modal and non-modal or only modal?

PR #410, opened 2019-06-21
Arrow nav

rwwagner90:
Related to #198

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -215,11 +215,19 @@ export class Step extends Evented {
    */
   _addKeyDownHandler(element) {
     const KEY_ESC = 27;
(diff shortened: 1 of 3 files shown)


src/js/step.js
```diff
@@ -215,12 +215,20 @@ export class Step extends Evented {
    */
   _addKeyDownHandler(element) {
     const KEY_ESC = 27;
+    const LEFT_ARROW = 37;
+    const RIGHT_ARROW = 39;
 
     element.addEventListener('keydown', (e) => {
-      switch(e.keyCode) {
+      switch (e.keyCode) {
```
chuckcarpenter: looks like this is the only change from previous PR, I realize you're building on those changes.

PR #445, opened 2019-07-17
allow changing the tippy and shepherd css classes and attributes pref…

genadis:
Similar to what tippy allows to do: https://github.com/atomiks/tippyjs/blob/v5/rollup.config.js#L9
Handy for compiling customised version and preventing possible collisions for complex applications environments.


rwwagner90:
Hi @genadis thanks for the PR! I am curious, did you run into a collision somewhere? We tried to prefix all classes with `shepherd-` to avoid that.

genadis:
Hi, thank you for this great project!
We have applications that work on sites we have no control of. So if they use shepherd as well, we would have a collision, so we need to use different prefix then shepherd.

genadis:
@rwwagner90 I've done styling fixes as well here: https://github.com/genadis/shepherd/compare/changable-prefixes...genadis:style-fixes
We prefer to use a darker header with white text and have the main text black with bright background. When I used this configuration, I've noticed some styling issues.

Would be happy to do a PR for the above, just need to know which branch to follow with this PR or the master. Hope it makes sense...

rwwagner90:
> We have applications that work on sites we have no control of. So if they use shepherd as well, we would have a collision, so we need to use different prefix than shepherd.

Ah, I did not think of that case. Would it be sufficient to add the prefix to the root `.shepherd` class? Since that wraps most of the other stuff, it might work?

As a side note, I would be very interested in showcasing  some of your Shepherd usages in our README, if you have some things to share 😃 

> We prefer to use a darker header with white text and have the main text black with bright background. When I used this configuration, I've noticed some styling issues.

Did you set the styles via overriding the scss variables or how exactly did you attempt this? I think this, and the other code you mentioned, should be another issue/PR, let's just focus on prefixes here.

genadis:
@rwwagner90 couple of reasons I can think of that adding prefix only to the root `.shepherd` won't work:
1. most of the styling is not wrapped by the root class so we might mess up the other shepherd styling on they ours.
2. if we use tippy with custom classes such as `ours-tippy` styling will be broken.
3. it is not only the css classes, shepherd adds data attributes to the body and uses ids as well that should be unique. if there is another shepherd on page there would be a collision.

I believe it's not a bad thing to have a single point where the classes names and attributes defined and used as constants, same for css. can potentially prevent typo errors.

* I've updated the branch to solve the conflicts after your recent commits

Regarding showcasing, we just started the integration of shepherd in our products, so once it goes to production will be in touch :)

rwwagner90:
> most of the styling is not wrapped by the root class so we might mess up the other shepherd styling or they ours.

Yeah, I was suggesting that we alter the styles to make everything nested under the root class.

> it is not only the css classes, shepherd adds data attributes to the body that should be unique. if there is another shepherd on page there would be a collision.

For sure. From looking at the code, it seems it relies on your building Shepherd yourself and replacing with rollup or am I misunderstanding? I think we should make the prefix a runtime option and append it to the outer class and the data attributes.

genadis:
Yeah, It relies on me building Shepherd. during build I set environmental variable/edit the rollup.
In addition it relies on editing the namespaces in `_variables.scss`: https://github.com/genadis/shepherd/blob/changable-prefixes/src/scss/_variables.scss#L3
In tippy they manage to inject scss variables from rollup, but unfortunately it didn't work for me.

Not sure how we can make the prefix a runtime option. for the attributes, sure, but how can we do it for the css generated from scss at build time? It could be great, but not sure there are enough use cases to be worth the effort, if somebody (like us :)) builds custom apps we don't mind having a branch with our custom build.
There are tools like: https://github.com/cssinjs/jss or a list here: http://michelebertoli.github.io/css-in-js/. But it will pretty much require a rewrite.

Regarding nesting the styles under the root class, not sure it covers all the cases. Consider the following:
App A - uses the standard build of shepherd.
App B (ours) - loaded on page of App B with prefix: `b-shepherd`.
The regular styles would be fine. But what if App A owner wants to style shepherd and does something like:
```css
.app-a .shepherd-header {
   background-color: red;
}
```
App A owner is not bound by the constraint "always use the shepherd root class to scope the styling". He can just add plain css to his site, we have no control over it.

On the other hand if we prefix everything the above will not effect App B custom built shepherd version which will have `b-shepherd-header` class.

rwwagner90:
@genadis Give me a couple days to play with this. Your solution may well be the best one, but I want to flesh out some ideas before merging.

genadis:
@rwwagner90 thank you for investing your time into this

rwwagner90:
@genadis I think maybe CSSinJS would work best for this. I started playing around with it and got it working for the `shepherd-header` class. Can you take a look and let me know your thoughts on this approach? https://github.com/shipshapecode/shepherd/pull/450

rwwagner90:
@genadis I was liking CSSinJS, so I spent the last couple days converting fully over to it, which will make this PR irrelevant, I am afraid. Please check out #450 when you have some time, and let me know if it will work for your needs. I still need to write up the guides for it, which I hope to do later today or tomorrow.

genadis:
@rwwagner90 seems you did :) 
I'll check it, thanks

rwwagner90:
@genadis I also just merged the changes to add the prefix in the data attributes https://github.com/shipshapecode/shepherd/pull/452

Is there anything else we need? Perhaps try out master and see if it works for you and let me know if we need any more changes?

PR #458, opened 2019-07-22
Convert to Preact components

rwwagner90:
This may or may not actually end up being merged, but the idea behind it is we can convert everything to Preact components, and no longer need to ship a million polyfills to support IE, if we convert things correctly. We'll see how it goes!


genadis:
@rwwagner90  why not something like https://stenciljs.com/ ? it's tiny fast, standard, loads polyfills dynamically, works across all frameworks, easy to start with, has great tooling, has all the bells and whistles.
We had a very good experience with it with our projects. In fact when we use shepherd we first wrap it in stencil web component.

rwwagner90:
@genadis I mainly went with preact because it was the smallest thing I had heard of so far. I've never heard of StencilJS, but it could be a viable option. I think much of the work of splitting into a more componentized architecture will overlap with any framework, so perhaps we can finish doing Preact, and then look into something else. If you wanted to work on an alternate branch using Stencil, and mimicking the work I am doing here, we could compare the two to see which is smaller and more performant 😃

PR #474, opened 2019-08-01
Add option to specify container element for the modal

genadis:
We don't always want to append to modal to the body, in some cases we need it to be child of other container element. (for steps we have appendTo tippy option).

### DIFF
--- a/src/js/tour.jsx
+++ b/src/js/tour.jsx
@@ -46,6 +46,8 @@ export class Tour extends Evented {
    * @param {boolean} options.useModalOverlay Whether or not steps should be placed above a darkened
    * modal overlay. If true, the overlay will create an opening around the target element so that it
    * can remain interactive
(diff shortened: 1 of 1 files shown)


src/js/tour.jsx
```diff
@@ -46,6 +46,8 @@ export class Tour extends Evented {
    * @param {boolean} options.useModalOverlay Whether or not steps should be placed above a darkened
    * modal overlay. If true, the overlay will create an opening around the target element so that it
    * can remain interactive
+   * @param {HTMLElement} options.modalContainer Optional container Element for the modal.
```
rwwagner90: Can you also document this here please?  https://github.com/shipshapecode/shepherd/blob/master/docs-src/tutorials/02-usage.md

rwwagner90: I will document it, no worries!

genadis: thanks!

PR #476, opened 2019-08-01
fix content border radius

genadis:
at the moment `shepherdElementBorderRadius` has no effect on when `attachTo` is used. It works only for modal.
Because the radius is applied only to header and footer. applying it on content as well, fixes it.

### DIFF
--- a/src/js/components/shepherd-content/styles.js
+++ b/src/js/components/shepherd-content/styles.js
@@ -4,7 +4,8 @@ export default function contentStyles(variables) {
       background: variables.shepherdTextBackground,
       fontSize: 'inherit',
       outline: 'none',
-      padding: 0
+      padding: 0,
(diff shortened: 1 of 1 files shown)


src/js/components/shepherd-content/styles.js
```diff
@@ -4,7 +4,8 @@ export default function contentStyles(variables) {
       background: variables.shepherdTextBackground,
       fontSize: 'inherit',
       outline: 'none',
-      padding: 0
+      padding: 0,
+      borderRadius: variables.shepherdElementBorderRadius
```
rwwagner90: Can you move this to below `background` please?  Trying to keep them in alphabetical order.

rwwagner90: Never mind, I will merge and cleanup

PR #494, opened 2019-08-09
Fix for the wrong type definition of StepOptions.scrollTo

### LINKED ISSUE #490: Wrong type definition for scrollTo
Hi,

the type definition for `scrollTo` is wrong: https://github.com/shipshapecode/shepherd/blob/master/src/types/step.d.ts#L118

It is defined as boolean, but in reality it accepts  `{ behavior: string, block?: string }|boolean`

moxival:
Simple fix for the wrong type definition of StepOptions.scrollTo property

Fixes #490

### DIFF
--- a/src/types/step.d.ts
+++ b/src/types/step.d.ts
@@ -115,7 +115,7 @@ declare namespace Step {
     /**
      * Should the element be scrolled to when this step is shown?
      */
-    scrollTo?: boolean;
+    scrollTo?: boolean | ScrollIntoViewOptions;
 
     /**
      * A function that lets you override the default scrollTo behavior and


src/types/step.d.ts
```diff
@@ -115,7 +115,7 @@ declare namespace Step {
     /**
      * Should the element be scrolled to when this step is shown?
      */
-    scrollTo?: boolean;
+    scrollTo?: boolean | ScrollIntoViewOptions;
```
rwwagner90: ScrollIntoViewOptions is not defined. We'll need to add it as well.

rwwagner90: Looks like this is built into TS, so should be okay.

PR #500, opened 2019-08-12
Add missing role="document"

knoobie:
I'm sorry if this is totally wrong.. 

Why? https://github.com/twbs/bootstrap/pull/29172

### DIFF
--- a/src/js/components/shepherd-content/index.jsx
+++ b/src/js/components/shepherd-content/index.jsx
@@ -12,6 +12,7 @@ export default class ShepherdContent extends Component {
 
     return <div
       className={styles.content.trim()}
+      role="document"
     >
       <ShepherdHeader
         labelId={labelId}


rwwagner90:
@knoobie can you say more on this? To my knowledge, we have all the roles we need already.

knoobie:
@rwwagner90 see this comment in the linked repository: https://github.com/twbs/bootstrap/pull/29172#issuecomment-516806609

rwwagner90:
@knoobie I see the comment, but I don't think it applies to us. Have you noticed an issue of some sort?

knoobie:
Sadly I don't have a mac / voiceover to test with. 

The comment 

> contents of modal dialogs may be inaccessible if the dialog element has an accessible name set by aria-label or aria-labelledby.

just matched perfectly to your dialog:

div **aria-describedby="welcome-description"** class="class-1 class-2 demo-shepherd-has-cancel-link demo-shepherd-element" data-demo-shepherd-step-id="welcome" **role="dialog"** tabindex="0"

Better safe than sorry - but I can understand you, when you don't wanna have it for now.

rwwagner90:
@knoobie I confirmed with an a11y expert, and this is not an issue here. It becomes an issue only if your application has set `role="document"` or `role="application"` elsewhere. With that in mind, I believe this is an application concern and not something we need here. If anyone experiences an issue, we can certainly revisit!

PR #501, opened 2019-08-12
fixing a11y issue by adding lang attribute to html element

MelSumner:
If merged, this PR would resolve the issue where the demo app doesn't have a lang attribute specified on the html element.


rwwagner90:
Good catch, thanks @MelSumner!

PR #514, opened 2019-08-14
Use requestAnimationFrame to position modal opening

rwwagner90:
This helps to mitigate some of the glitchy scrolling where the modal overlay opening would visible jump around on scroll. It does not totally fix the problem, but is much better.

A possible way to improve things further would be to position with CSS transforms, rather than changing x and y.


rwwagner90:
@dipiana This should have fixed things a good bit. I think the one improvement we could still make, is to just do `getBoundingClientRect` once, then instead of setting x and y over and over again, we should reposition with CSS transforms based on the scroll position or something. Something like https://codepen.io/rachsmith/post/how-to-move-elements-on-scroll-in-a-way-that-doesn-t-suck-too-bad

If you want to give it a try, let me know! Otherwise, I think we're in a lot better place after this PR.

dipiana:
@rwwagner90, I am trying to understand the code at the moment, javascript is not my strength but let's see.
Could you please explain what "rafLoop" is for?

Thanks.

rwwagner90:
@dipiana it's for `requestAnimationFrame` so it will run each time `requestAnimationFrame` runs. I wouldn't worry too much about it, I think performance is pretty decent now, unless you are noticing big problems?

PR #526, opened 2019-08-22
Add includeStyles option

### LINKED ISSUE #497: Disable built in component styles
After upgrading our application to 4.6 (from 3.x) - I noticed that you have changed your styling approach from css files to js included css. Appreciate that! 

It just comes with the little downside, when you explicit wanna exclude some css. We removed all the button styling before - so your default styling won't interfere with bootstrap's styling.

Example:
````js 
buttons: [
    {
      action() {
        return this.cancel();
      },
      classes: 'btn btn-secondary',
      text: 'Beenden'
    }, {
      action() {
        return this.next();
      },
      classes: 'btn btn-primary',
      text: 'Weiter'
    }
  ]
````

Now your classes overwrite the bootstrap styling because of your inline css. Is there a way to disable for example the css of the buttons? Like "removeClasses" or "disableStyling".

Proposal (sorry, no js guy)

grafik

### PR DESCRIPTION
**Pull request by rwwagner90**
This will allow users who find themselves overriding a lot of styles to opt out of our base styles. Closes #497

### DIFF
--- a/src/types/step.d.ts
+++ b/src/types/step.d.ts
@@ -113,9 +113,9 @@ declare namespace Step {
     id?: string;
 
     /**
-     * Extra [options to pass to tippy.js]{@link https://atomiks.github.io/tippyjs/#all-options}
+     * An amount of padding to add around the modal overlay opening
      */
-    tippyOptions?: object;
+    modalOverlayOpeningPadding?: number;
 
     /**
      * Should the element be scrolled to when this step is shown?
@@ -144,6 +144,11 @@ declare namespace Step {
      * ```
      */
     text?: string | ReadonlyArray<string> | HTMLElement | (() => string | ReadonlyArray<string> | HTMLElement)
+
+    /**
+     * Extra [options to pass to tippy.js]{@link https://atomiks.github.io/tippyjs/#all-options}
+     */
+    tippyOptions?: object;
 
     /**
      * The step's title. It becomes an `h3` at the top of the step.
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by rwwagner90** [comment]
@knoobie @chadarsenault does this approach work for you guys? It basically removes most of the styles if you set `includeStyles: false`. It only keeps the modal and cancel link stuff, but header, footer, button, text, and content styles are removed.

File: src/types/step.d.ts [hunk 1]
```diff
@@ -145,6 +145,11 @@ declare namespace Step {
      */
     text?: string | ReadonlyArray<string> | HTMLElement | (() => string | ReadonlyArray<string> | HTMLElement)
 
+    /**
+     * Extra [options to pass to tippy.js]{@link https://atomiks.github.io/tippyjs/#all-options}
+     */
+    tippyOptions?: object;
```
chuckcarpenter: I like the alphabetizing

knoobie:
Thank you! The idea looks good! Have to find proper time to test it. (sorry) 

Proposal / idea for future improvement (dunno how much work that would be) - instead of one boolean:

```js
  styling: {
    header: true, 
    footer: true,
    modal: true,
    button: true,
    content: true,
    text: true
  }
```

rwwagner90:
@knoobie I don't think we should make it that granular. If you want to override a couple things we provide the variables to change the styles or overriding via CSS is always an option.

msteel9999:
To avoid wasting time as I did, note this isn't part of 9.1 release. Look at the source in master will show the includeStyles option is missing. Some error in the merge I think

rwwagner90:
@msteel9999 this was intentionally removed. Since the styles live in a CSS file now, which you must manually include, if you don't want the styles at all, you can simply not include that file.

rwwagner90:
@iPurpl3x the CSS is entirely optional. If you don't include it, the styles won't be there.

iPurpl3x:
Well yes, but I cannot selectively include CSS, like I would need all the layout CSS, but no colors etc... Anyway, I have now overridden the part that was conflicting with my button styling, so it's ok...

PR #530, opened 2019-08-31
remove rarely used get-value dependency

genadis:
the get-value is used 3 times and for trivial if checks.
removing it saves ~1K, most likely better performance (since simple if is used instead of get-value generic logic).
And finally less prone for errors, using strings for getting properties is not safe and cancels any autocompletion and typing.

(no new/change in functionality, so existing tests should cover the fix)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "element-matches": "^0.1.2",
     "es6-object-assign": "^1.1.0",
     "es6-symbol": "^3.1.1",
-    "get-value": "^3.0.1",
     "nano-css": "^5.2.0",
     "polished": "^3.4.1",
     "preact": "^8.5.2",
--- a/src/js/step.jsx
+++ b/src/js/step.jsx
@@ -1,4 +1,3 @@
-import get from 'get-value';
 import preact from 'preact';
 
 import { Evented } from './evented';
@@ -229,12 +228,11 @@ export class Step extends Evented {
    * @private
    */
   _createTooltipContent() {
-    const cancelIconEnabled = get(this, 'options.cancelIcon.enabled');
     let classes = this.options.classes || '';
     const descriptionId = `${this.id}-description`;
     const labelId = `${this.id}-label`;
 
-    if (cancelIconEnabled) {
+    if (this.options.cancelIcon && this.options.cancelIcon.enabled) {
       classes += ` ${this.classPrefix}shepherd-has-cancel-icon`;
     }
 
--- a/src/js/utils/tippy-popper-options.js
+++ b/src/js/utils/tippy-popper-options.js
@@ -1,5 +1,3 @@
-import get from 'get-value';
-
 const addHasTitleClass = (step) => {
   return { addHasTitleClass: _createClassModifier(`${step.classPrefix}shepherd-has-title`) };
 };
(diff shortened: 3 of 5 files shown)


rwwagner90:
I just changed this to use `get-value` I am not interested in removing it.

genadis:
@rwwagner90 What is the added value of using `get-value`?
do you not agree with the disadvantages I've listed?

genadis:
@rwwagner90 allow me to note: It's your project, so your decision of how to handle pull requests, issues and what to include in project.

But closing pull requests, that contributors put effort in, and include reasoning behind it, without providing any explanation and complete disregard.. Really discourages from future collaboration and contribution.

rwwagner90:
@genadis Apologies for not explaining the reasoning in depth, but if you checked the commit history, you would see I literally just added `get-value`, and the code before adding `get-value` was essentially identical to what you were proposing here.

I specifically added it to avoid the need for `foo && foo.bar && foo.bar.baz && foo.bar.baz.one && foo.bar.baz.one.two`. 

Without `get-value` the min + gzip bundle size is 29.67kb and with it, it is 29.86kb, so removing it only saves 0.19kb. I do not think that very tiny amount of bytes is worth removing the package personally.

genadis:
@rwwagner90 I guess it goes down to style and best practices preferences.

Ours: As a general rule of thumb, use third party dependencies only if it adds value, saves time and reduced chance for errors. Size is a factor too, but not the only one. In this case, I personally don't see value in adding third party code for this use case. If there were lots of usage for that I might agree with you, but there is not. In addition, I'm personally agains using string accessors which are totally unsafe (in the modern es6/typescript).

I can't argue with personal preference, so we'll just maintain our own branch for such fixes.

Thank you for your efforts in this great project!

rwwagner90:
@genadis you really need to fork for this tiny issue?

genadis:
@rwwagner90 Unfortunately yes, we choose carefully third party dependencies, and try to keep to a minimum.

But now worries... It's a small change, so we can manage it and still keep our branch close to the main project.

rwwagner90:
@genadis can you please explain? Seems really silly to fork for 0.19kb.

genadis:
@rwwagner90 Basically the reasoning is what I've already mentioned, for us, it's not only a matter of package size. Just look at dependencies of https://www.npmjs.com/package/tippy.js and https://www.npmjs.com/package/popper.js. there are none...

> Ours: As a general rule of thumb, use third party dependencies only if it adds value, saves time and reduced chance for errors. Size is a factor too, but not the only one. In this case, I personally don't see value in adding third party code for this use case. If there were lots of usage for that I might agree with you, but there is not. In addition, I'm personally agains using string accessors which are totally unsafe (in the modern es6/typescript).

rwwagner90:
@genadis I just don't understand. It's a super small dependency, used by almost 3 million projects and downloaded 9,616,175 times per week. I could understand if I added something large, insecure, or error prone, but this is very straightforward, to save us from typing super long checks every time we want to access something nested.

Our goal is not to have zero dependencies here. Obviously, we want to keep things as minimal as possible, but when dependencies help save me some time, I tend to like to use them. There are some much bigger things we are working on to help reduce size, like removing `polish` and having users set all the style variables, instead of auto calculating, and we will soon remove all the IE11 polyfills, in favor of listing instructions for including the polyfills in apps, if you need to support IE11. This will reduce the size significantly.

genadis:
Your effort to remove bigger stuff is great.

> Our goal is not to have zero dependencies here. Obviously, we want to keep things as minimal as possible, but when dependencies help save me some time, I tend to like to use them.

To tell you the truth, I'm not sure how manually typing string like: `'tour.options.styleVariables.shepherdElementZIndex'` saves time instead of using editor's autocomplete. What if you make a mistake in the string?
There is no difference in the amount of the code, no nested ifs, there are no complex checks, it's just basic options parsing that is done few times, so I don't see the value in using it.

I know the package is popular, but everything has it's purpose... If there were complex stuff or some dynamic object manipulation and lots of usage that's a different story. I'm personally pro type safety (use lots of typescript/es6/flow) especially when it comes to things like options.

rwwagner90:
@genadis we're trying to remove as many deps as possible for v6, so I will accept this. It has conflicts, so I am going to manually do the changes again.

PR #531, opened 2019-09-01
Refactor styleVariables, remove automatic calculations

rwwagner90:
This removes polished as a dependency, changes the variable names, and requires the user to explicitly define all styles for each state. We were calculating sane defaults, based on other variables you passed, but that was buggy, so setting them all is best.

This will save us ~4 kb from our bundle size as well.


docs-src/tutorials/03-styling.md
```diff
@@ -9,55 +11,31 @@ let tour = new Shepherd.Tour({
...
-| shepherdThemeTextSecondary | The text color for the secondary button                     | transparentize(0.25, readableColor(shepherdThemeSecondary))
-| shepherdThemeTextHeader    | The text color for the header                               | transparentize(0.25, readableColor(shepherdHeaderBackground))
-| shepherdThemeTextColor     | The text color for the step content                         | transparentize(0.25, readableColor(shepherdTextBackground))
+If you have a situation where you are running two Shepherd instances on a page, and they need to be styled
+differently, you may want to prefix the class names. This is now possible with the `classPrefix` option.
 
-We use [polished](https://github.com/styled-components/polished) to calculate lighter, darker, and the most readable colors.
+```js
+const tour = new Shepherd.Tour({
+  classPrefix: 'my-tour-'
```
chuckcarpenter: 👍

PR #532, opened 2019-09-02
Fix tab focus when there are no focusable elements

genadis:
The issue:
if a step contains only 1 focusable element clicking on TAB does not focus on it, the focus stays on the step element.

This fix will focus on the first element and keep the focus on it for subsequent tabs.

If I'm not mistaken, to add cypress a11 test for it, we need to change the welcome tour, since there is no step with single focusable element. Not sure you want to do it, so leave it to you.

[The second commit](https://github.com/shipshapecode/shepherd/pull/532/commits/7479441b0e8f2d4b326d6448430795de19cbbb37) handles another related issue: If there are no focusable elements, pressing TAB will move the focus outside of the tour.
I'm guessing we want to keep it inside. @rwwagner90  What do you think?

### DIFF
--- a/src/js/components/shepherd-element/index.jsx
+++ b/src/js/components/shepherd-element/index.jsx
@@ -69,11 +69,6 @@ export default class ShepherdElement extends Component {
           e.preventDefault();
           break;
         }
-        if (this.focusableElements.length === 1) {
-          e.preventDefault();
-          this.firstFocusableElement.focus();
-          break;
-        }
         // Backward tab
         if (e.shiftKey) {
           if (document.activeElement === this.firstFocusableElement) {
--- a/src/js/components/shepherd-element/index.jsx
+++ b/src/js/components/shepherd-element/index.jsx
@@ -67,6 +67,7 @@ export default class ShepherdElement extends Component {
       case KEY_TAB:
         if (this.focusableElements.length === 1) {
           e.preventDefault();
+          this.firstFocusableElement.focus();
           break;
         }
         // Backward tab
--- a/src/js/components/shepherd-element/index.jsx
+++ b/src/js/components/shepherd-element/index.jsx
@@ -65,6 +65,10 @@ export default class ShepherdElement extends Component {
     const { tour } = this.step;
     switch (e.keyCode) {
       case KEY_TAB:
+        if (this.focusableElements.length === 0) {
+          e.preventDefault();
+          break;
+        }
         if (this.focusableElements.length === 1) {
           e.preventDefault();
           this.firstFocusableElement.focus();


rwwagner90:
@genadis good catch here! 

I think the issue lies in my incorrect query for focusable elements. https://github.com/shipshapecode/shepherd/blob/master/src/js/components/shepherd-element/index.jsx#L22

I believe the list of focusable elements should include `this.element` then, the code should work as is I think. Thoughts? Then there would always be one focusable element.

genadis:
@rwwagner90 I guess it depends on the UX, if we want user to be able to circulate between the inner elements or to include the step element itself in the circulation.

In case we decide to include the step element in the focusable elements list, then I believe you are right, there won't be need for this pull request.

src/js/components/shepherd-element/index.jsx
```diff
@@ -65,8 +65,13 @@ export default class ShepherdElement extends Component {
     const { tour } = this.step;
     switch (e.keyCode) {
       case KEY_TAB:
+        if (this.focusableElements.length === 0) {
+          e.preventDefault();
+          break;
+        }
         if (this.focusableElements.length === 1) {
```
rwwagner90: Could we remove this case? Essentially just have the base case of no focusable elements, and otherwise do the normal tabbing. I believe it should work without having a case for a single element. Do you mind trying it?

Also, I am curious how would you end up with a tour element with no focusable elements? Are the controls for advancing the tour outside of the tooltip?

genadis: > Could we remove this case? Essentially just have the base case of no focusable elements, and otherwise do the normal tabbing. I believe it should work without having a case for a single element. Do you mind trying it?

This is for the case there are no focusable elements (tried it before the commit). what happens is when user clicks TAB, the focus leaves the tour and moves outside the step.

> Also, I am curious how would you end up with a tour element with no focusable elements? Are the controls for advancing the tour outside of the tooltip?

We have a step that we ask user to perform an action ("Click here to ...."). once he clicks, we advance to next step programatically. The step is minimal without option to exit the tour and with no buttons.
(we rarely do it but there are a few cases that are critical so we require users to take action)

genadis: Regarding accessibility we are now in brainstorming of how to solve it actually. Since the step requires user to perform action, to perform it user needs to have focus on the clickable element, and there is no such support it shepherd for that.

rwwagner90: > This is for the case there are no focusable elements (tried it before the commit). what happens is when user clicks TAB, the focus leaves the tour and moves outside the step.

I'm not sure we are on the same page here. I am saying to keep the no focusable elements case, which is the `0` case, and remove the `1` case in favor of the default functionality. It should work.

> Regarding accessibility we are now in brainstorming of how to solve it actually. Since the step requires user to perform action, to perform it user needs to have focus on the clickable element, and there is no such support it shepherd for that.

This is a separate issue. We should probably allow tabbing out of the tour, if `useModalOverlay` is false, but that would be a separate thing from fixing this.

rwwagner90:
@genadis to be clear, this is what I suggest:

```js
if (this.focusableElements.length === 0) {
          e.preventDefault();
          break;
        }
        // Backward tab
        if (e.shiftKey) {
          if (document.activeElement === this.firstFocusableElement) {
            e.preventDefault();
            this.lastFocusableElement.focus();
          }
        } else {
          if (document.activeElement === this.lastFocusableElement) {
            e.preventDefault();
            this.firstFocusableElement.focus();
          }
        }
```

If there are no focusable elements, it will hit the first `0` case. Otherwise, it should  work like normal. Is there a problem with this?

genadis:
I see what you mean, checked it (all works), and removed the `this.focusableElements.length === 1` case.

Thanks

src/js/components/shepherd-element/index.jsx
genadis: @rwwagner90 agree, it's a separate issue.

as a side note: since the step is a dialog if I'm not mistaken, according to accessibility standards it should not allow tabbing out, regardless if there is an overlay.

rwwagner90: @genadis yes, that is why it was originally set to not allow tabbing out at all. I'm open to suggestions on how to make things work for both of us, while still adhering to a11y standards.

genadis: @rwwagner90 how do you feel about adding an event that fires if user activates a step element - `space` or `enter` when the focus is on the step element.

We can then leverage this event in our app:
1. in our step that requires the user to click without the next button, we will add hidden accessibility message asking user to use the space or enter to move to next step.
2. catch the event, do the action programatically and proceed to next step.

what do you think?

genadis: actually we can catch the keboard event in our app directly as well, so no need.

thanks

rwwagner90: I'm not sure I quite follow, but if you can handle it in your app, that is great! Please let me know if you end up needing anything on the Shepherd end.

PR #554, opened 2019-09-10
Replace uniqueId with uuid

### LINKED ISSUE #550: Reuse uniqueId logic

rwwagner90:
Closes #550 

@chuckcarpenter do you foresee any issues with replacing the old `step-1`, `step-2` style ids with uuid? The thought here is we can reduce code, and use uuid for all ids, as well as guarantee ids are unique.


chuckcarpenter:
The use case where this was used, can be met with the `classes` prop on a step. So this should be good.

PR #560, opened 2019-09-13
Add tests for multiple tours

### LINKED ISSUE #548: Add tests for multiple tours on a page
This will allow us to be certain you can have two tours, as long as only one is running at a time.

Things to test:

- [x] It is possible to style the tours independently 
- [x] The `data-` attributes are prefixed, so they do not interfere with each other
- [x] The tour active classes, etc are prefixed, so it knows which tour to interact with
- [x] The modal overlays only show one overlay, even though two exist on the page

chuckcarpenter:
Closes #548 

This branch removes the need for body data attributes and tests to ensure tours can be run independently.

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -199,13 +199,24 @@ export class Step extends Evented {
   }
 
   /**
+   * _concatClasses concatonates default step classes to step specific classes
+   * @param {String} classString The string of classes for the step
+   * @private
+   */
+  _concatClasses(classString) {
+    const tourOptions = this.tour && this.tour.options && this.tour.options.defaultStepOptions;
+
+    return tourOptions.classes.concat(` ${classString}`);
+  }
+
+  /**
    * Creates Shepherd element for step based on options
    *
    * @return {Element} The DOM element for the step tooltip
    * @private
    */
   _createTooltipContent() {
-    let classes = this.options.classes || '';
+    const classes = this.options.classes || '';
     const descriptionId = `${this.id}-description`;
     const labelId = `${this.id}-label`;
 
@@ -249,14 +260,19 @@ export class Step extends Evented {
    * @private
    */
   _setOptions(options = {}) {
+    const tourOptions = this.tour && this.tour.options && this.tour.options.defaultStepOptions;
     this.options = Object.assign(
       {
         arrow: true
       },
-      this.tour && this.tour.options && this.tour.options.defaultStepOptions,
+      tourOptions,
       options);
 
     const { when } = this.options;
+
+    if (tourOptions.classes) {
+      this.options.classes = this._concatClasses(this.options.classes);
+    }
 
     this.destroy();
     this.id = this.options.id || `step-${uuid()}`;
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -199,17 +199,6 @@ export class Step extends Evented {
   }
 
   /**
-   * _concatClasses concatonates default step classes to step specific classes
-   * @param {String} classString The string of classes for the step
-   * @private
-   */
-  _concatClasses(classString) {
-    const tourOptions = this.tour && this.tour.options && this.tour.options.defaultStepOptions;
-
-    return tourOptions.classes.concat(` ${classString}`);
-  }
-
-  /**
    * Creates Shepherd element for step based on options
    *
    * @return {Element} The DOM element for the step tooltip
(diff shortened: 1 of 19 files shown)


test/cypress/integration/element-targeting.spec.js
```diff
@@ -9,7 +9,7 @@ describe('Attaching tooltips to target elements in the DOM on each step', () =>
     cy.visit('/test/dummy/', {
       onLoad(contentWindow) {
         if (contentWindow.Shepherd) {
-          return Shepherd = contentWindow.Shepherd;
+          return (Shepherd = contentWindow.Shepherd);
```
rwwagner90: I'm curious, what are the parentheses for?

chuckcarpenter: I think this was added from the linter

test/cypress/integration/element-targeting.spec.js
```diff
@@ -43,4 +43,46 @@ describe('Attaching tooltips to target elements in the DOM on each step', () =>
...
+
+    afterEach(() => {
+      firstTour.complete();
+      secondTour.complete();
+    });
+    it('applies default classes only on each individual tour', async function() {
+      firstTour.start();
+      secondTour.start();
+
+      cy.get('.shepherd-element').and(result => {
```
rwwagner90: I think we can simplify these and do like:

```js
cy.get('.shepherd-element').should('have.length', 2);
```

src/js/step.js
```diff
@@ -2,7 +2,12 @@ import { Evented } from './evented.js';
 import autoBind from './utils/auto-bind.js';
 import { isElement, isFunction, isUndefined } from './utils/type-check.js';
 import { bindAdvance } from './utils/bind.js';
-import { setupTooltip, parseAttachTo, normalizePrefix, uuid } from './utils/general.js';
+import {
+  setupTooltip,
```
rwwagner90: It looks like we still have formatting changes here

src/js/step.js
```diff
@@ -97,7 +102,9 @@ export class Step extends Evented {
   constructor(tour, options = {}) {
     super(tour, options);
     this.tour = tour;
-    this.classPrefix = this.tour.options ? normalizePrefix(this.tour.options.classPrefix) : '';
+    this.classPrefix = this.tour.options
+      ? normalizePrefix(this.tour.options.classPrefix)
```
rwwagner90: And here

src/js/step.js
```diff
@@ -205,21 +210,20 @@ export class Step extends Evented {
...
    */
   _createTooltipContent() {
-    let classes = this.options.classes || '';
+    const classes = this.options.classes || '';
     const descriptionId = `${this.id}-description`;
     const labelId = `${this.id}-label`;
 
     const ShepherdElementComponent = new ShepherdElement({
       target: document.body,
-      props:
```
rwwagner90: And here etc

src/js/tour.js
```diff
@@ -64,10 +64,10 @@ export class Tour extends Evented {
     this.modal = new ShepherdModal({
       target: options.modalContainer || document.body,
       props:
-        {
-          classPrefix: this.classPrefix,
-          styles: this.styles
-        }
+      {
```
chuckcarpenter: this is eslint, you'll have to deal @rwwagner90

test/unit/tour.spec.js
```diff
@@ -21,14 +21,14 @@ describe('Tour | Top-Level Class', function() {
     instance.complete();
   });
 
-  describe('constructor', function() {
-    it('creates a new tour instance', function() {
+  describe('constructor', function () {
```
chuckcarpenter: @rwwagner90 should also stay

src/js/step.js
chuckcarpenter: the spacing is an eslint thing

test/unit/tour.spec.js
```diff
@@ -120,37 +120,29 @@ describe('Tour | Top-Level Class', function() {
...
-    describe('.start()', function() {
-      it('starts a tour that is the current active', function() {
+    describe('.start()', function () {
+      it('starts a tour that is the current active', function () {
         instance.start();
 
         expect(instance).toBe(Shepherd.activeTour);
       });
-
-      it('adds the current tour\'s "id" property to the body as a data attribute', function() {
```
chuckcarpenter: removing this is the only functional change

PR #581, opened 2019-09-26
Implement ability to update step options

VladimirAnaniev:
I have implemented the changes discussed in #563

### DIFF
--- a/src/js/components/shepherd-text.svelte
+++ b/src/js/components/shepherd-text.svelte
@@ -1,10 +1,10 @@
 <script>
-  import { onMount } from 'svelte';
+  import { afterUpdate } from 'svelte';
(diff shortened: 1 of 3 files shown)


rwwagner90:
This is awesome! Thanks so much for the PR 😃

Steven-Wagner:
I know this was from a while ago but does anyone know if this fix will have an effect on updating popperOptions? I have a shepherd modal inside an iframe and need to force the modal to change its popper offset after a scrolling event outside the iframe. I have tried using `updateStepOptions()` and it seems to be updating the step in theory, but it's not rendering the changes to the screen. I am able to get it to work by hiding and then reshowing the modal, so maybe that's the best I can hope for.

PR #589, opened 2019-09-30
Adjust modal opening according to scrollParent

### LINKED ISSUE #585: Modal opening overflow when target is in a scrollable area
## The issue
When the target element is in a scrollable area and is overflowing, the highlight does not take into account that a part of the element is not visible. The popover position alsodoes not go outside the scrollable parent even if there is available space.

Issue demonstrated in Stackblitz: https://stackblitz.com/edit/shepherd-target-scrollable?file=index.html

## Proposed solution for highlight overflowing
On show of the step we can get the scrollable parent if there is one:  
```typescript
function getScrollParent(element: HTMLElement): HTMLElement {
      if (!element) {
         return null;
      }

      const overflowY = window.getComputedStyle(element).overflowY;
      const isScrollable = overflowY !== 'hidden' && overflowY !== 'visible';

      if (isScrollable && element.scrollHeight >= element.clientHeight) {
         return element;
      }

      return getScrollParent(element.parentElement) || document.body;
   }
```
Then in the render loop in `shepherd-modal.svelte` we can compare the scroll parent's position and our target position like so:
```typescript
function getVisibleHeight(element: HTMLElement, scrollParent: HTMLElement): {top: number, height: number} {
      const elementTop = element.getBoundingClientRect().top;
      const elementBottom = element.getBoundingClientRect().bottom;

      const scrollTop = scrollParent.getBoundingClientRect().top;
      const scrollBottom = scrollParent.getBoundingClientRect().bottom;

      const top = Math.max(elementTop, scrollTop);
      const bottom = Math.min(elementBottom, scrollBottom);
      const height = bottom - top;

      return {top, height};
   }
```
For a horisontal scroll the implementation should be analogous. I can open a PR about this in the coming days.

## Popover position
This might be an issue in the popper library. Will Tether handle this use case correctly? Is there a solution for this issue in the meantime?

VladimirAnaniev:
Fixes #585. 

Example after applying the fix:
scroll-overflow html

### DIFF
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -32,16 +32,18 @@
   /**
    * Uses the bounds of the element we want the opening overtop of to set the dimensions of the opening and position it
    * @param {HTMLElement} targetElement The element the opening will expose
+   * @param {HTMLElement} scrollParent The scrollable parent of the target element
    * @param {Number} modalOverlayOpeningPadding An amount of padding to add around the modal overlay opening
    */
-  export function positionModalOpening(targetElement, modalOverlayOpeningPadding = 0) {
+  export function positionModalOpening(targetElement, scrollParent, modalOverlayOpeningPadding = 0) {
     if (targetElement.getBoundingClientRect) {
-      const { x, y, width, height, left, top } = targetElement.getBoundingClientRect();
+      const { y, height } = _getVisibleHeight(targetElement, scrollParent);
+      const { x, width, left } = targetElement.getBoundingClientRect();
 
       // getBoundingClientRect is not consistent. Some browsers use x and y, while others use left and top
       openingProperties = {
         x: (x || left) - modalOverlayOpeningPadding,
-        y: (y || top) - modalOverlayOpeningPadding,
+        y: y - modalOverlayOpeningPadding,
         width: (width + (modalOverlayOpeningPadding * 2)),
         height: (height + (modalOverlayOpeningPadding * 2))
       };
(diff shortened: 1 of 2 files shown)


rwwagner90:
@VladimirAnaniev I just tested this out and it causes regressions on both the landing page and the demo site. If you run `yarn watch` you can see the opening does not go around the all the boxes fully and if you open the index.html in the root to see the landing page, you can see most elements are not highlighted at all.

I am going to have to revert the changes for now, until we can figure out a fix.

rwwagner90:
@VladimirAnaniev please let me know if there is anything I can do to help here. I do think supporting your use case would be great, but we need to make sure the regressions do not occur.

VladimirAnaniev:
@rwwagner90 Thank you, I can take a look again tonight or tomorrow. 

It would be good to cover this case with a test to avoid regressions 🙂. Do you think a unit test can suffice?

rwwagner90:
@VladimirAnaniev I am not sure the best way to test. I am okay with any type of test that ensures the openings are the right size for the various usages.

VladimirAnaniev:
@rwwagner90 I just opened the demo website and found that the body's height is set to 100%:
Inkedshepherd-demo-body

With the new logic this would mean that the targeted element is not visible if it is below. If the `height: 100%` style is removed it should work fine. Seems like a layout thing in this case.

rwwagner90:
@VladimirAnaniev we should support either. The existing code uses the normal `getBoundingClientRect` and returns the right coordinates. The changes should continue to return the right bounding box for all existing cases, and fix the cases you need it to fix. Try running `yarn watch` on your branch. Click next a few times and see the bounding box is not correct.

PR #595, opened 2019-10-02
Handle modal opening for targets inside scrollable area

### LINKED ISSUE #585: Modal opening overflow when target is in a scrollable area

VladimirAnaniev:
Resolves #585, fixes issue from #589 where elements without scroll parent were clipped when they exceeded the height of the body.

### DIFF
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -140,14 +140,16 @@
   * @private
   */
   function _getScrollParent(element) {
+    if (!element) {
+      return null;
+    }
+
     const isHtmlElement = element instanceof HTMLElement;
     const overflowY = isHtmlElement && window.getComputedStyle(element).overflowY;
     const isScrollable = overflowY !== 'hidden' && overflowY !== 'visible';
 
 
-    if (!element) {
-       return document.body;
-    } if (isScrollable && element.scrollHeight >= element.clientHeight) {
+    if (isScrollable && element.scrollHeight >= element.clientHeight) {
        return element;
     }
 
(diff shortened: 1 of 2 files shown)


rwwagner90:
Thanks for fixing these issues @VladimirAnaniev! This is a great addition to the library 😃

PR #596, opened 2019-10-02
Feature/buttons disabled option

### LINKED ISSUE #582: Disabled option for buttons
It would be helpful if the StepOptionsButton had a disabled property. This property could also be a function to allow more complex logic.
```typescript
disabled: boolean | (() => boolean);
```

If the property is true, the button would have the `disabled` property to true.

VladimirAnaniev:
Closes #582  
**Screenshot of a disabled button**  
Screenshot from 2019-10-02 22-22-34


rwwagner90:
@VladimirAnaniev thanks for the PR! If you haven't already, we are running a contest for giving away some Ember Temperature Control Mugs for Hacktoberfest, if you would like to enter. Opening a PR for a Hacktoberfest issue gives you 100 entries! https://twitter.com/shipshapecode/status/1179190723666759681

src/types/step.d.ts
```diff
@@ -216,6 +216,12 @@ declare namespace Step {
      * The HTML text of the button
      */
     text?: string;
+
+    /**
+     * Whether the button should be disabled
+     * When the value is `true` the button will be disabled
+     */
+    disabled?: boolean;
```
rwwagner90: Didn't we want to allow passing functions for disabled as well?

VladimirAnaniev: I thought about it but was not sure when we should re-evaluate the function.  
When do you think is suitable? On every show event? On an interval?

rwwagner90: I would think we would do the same thing we do with `text`.

test/cypress/integration/test.acceptance.js
```diff
@@ -134,6 +134,56 @@ describe('Shepherd Acceptance Tests', () => {
...
+              {
+                classes: 'enabled-button',
+                text: 'enabled'
+              }
+            ]
+          }
+        ]);
+        tour.start();
+        tour.show('test-step');
+        cy.get('.shepherd-button.disabled-button')
```
rwwagner90: Would we be able to move these tests to a unit test of the shepherd-button component? They are fairly simple, and unit tests will add to coverage.

VladimirAnaniev: Yes, this makes sense, I have moved them :)

rwwagner90:
Thanks so much for the PR @VladimirAnaniev! Really appreciate your help with this feature 😃

PR #601, opened 2019-10-05
Update step.js

tacoelho:
Disable buttons.


rwwagner90:
@tacoelho This PR appears to only be adding a line in the documentation. Did you intend for their to be more changes? We already have https://github.com/shipshapecode/shepherd/pull/596 for disabling buttons

tacoelho:
> @tacoelho This PR appears to only be adding a line in the documentation. Did you intend for their to be more changes? We already have #596 for disabling buttons

Yes, I was going to work more on this. You can close my PR, my bad, I didn’t see that.

PR #603, opened 2019-10-05
Add step.tetherOptions param

### LINKED ISSUE #594: Add step.tetherOptions
This should merge with the default options and allow overriding them to set your own options to pass to Tether

lex111:
Resolves #594 

@rwwagner90 This works, but I don’t think I did it right. Moreover, I do not know how to check types definitions. But I want to look into the matter, so please help.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@babel/preset-env": "^7.6.0",
     "@testing-library/jest-dom": "^4.1.0",
     "@testing-library/svelte": "^1.8.0",
+    "@types/tether": "^1.4.4",
     "autoprefixer": "^9.6.1",
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^24.9.0",
--- a/src/js/utils/general.js
+++ b/src/js/utils/general.js
@@ -97,7 +97,8 @@ export function uuid() {
  * @private
  */
 function _makeTooltipInstance(attachToOptions, step) {
-  let tetherOptions = Object.assign({}, {
+  const { constraints, optimizations, ...options } = step.tour.options.tetherOptions;
+  const defaultTetherOptions = {
     classPrefix: 'shepherd',
     constraints: [
       {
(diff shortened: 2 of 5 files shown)


src/js/utils/general.js
```diff
@@ -97,7 +97,7 @@ export function uuid() {
  * @private
  */
 function _makeTooltipInstance(attachToOptions, step) {
-  let tetherOptions = {
+  let tetherOptions = Object.assign({}, {
```
rwwagner90: The problem with this is `Object.assign` only works one level deep, so anything where an object is used for the property could have issues merging different keys inside that option. For example:

```js
new Tether({
  element: yellowBox,
  target: greenBox,
  attachment: 'top left',
  targetAttachment: 'bottom left',
  optimizations: {
    moveElement: false
  }
});
```

I think for anything that is a hash, like `optimizations` or `classes` we want to manually add those in, to ensure we can merge the nested options and overwrite where necessary.

Does that make sense? We had something similar for modifiers in the old `popperOptions`, if you take a look at the history.

package.json
```diff
@@ -56,6 +56,7 @@
     "@babel/preset-env": "^7.6.0",
     "@testing-library/jest-dom": "^4.1.0",
     "@testing-library/svelte": "^1.8.0",
+    "@types/tether": "^1.4.4",
```
rwwagner90: I didn't even know we had types for tether! We probably need to update this.

lex111: That's it! But how to use them correctly, I don’t know, actually. I mean, now in the Rollup configuration there isn’t even a TS code processing plugin at all.

src/js/utils/general.js
lex111: Oopsy, I didn’t know about this, maybe it makes sense to use the lodash function just for this case!

-> https://lodash.com/docs#merge

rwwagner90: @lex111 we don't want to add lodash here.

I would recommend something like:

```js
  tetherOptions.optimizations = {
    ...tetherOptions.optimizations,
    ...step.options.tetherOptions.optimizations,
  };
```

This is just an example, and I have not tried it.

lex111: Got you, but we don’t need to import the whole library, if I’m not mistaken, its functions can be used separately (as packages), so this will not increase the bundle size.

rwwagner90: They can be used separately, but we do not want to add them

lex111: @rwwagner90 sorry for the long wait, I did it, please take a look.

lex111:
:(

rwwagner90:
@lex111 I opened a new PR to add this. This PR did not do what was intended. Thanks for the effort though!

PR #604, opened 2019-10-05
Add option to set aria-label on step buttons

### LINKED ISSUE #540: Add option to set aria-label on step buttons
Need to add a label property in button configurations in steps config just like you have provided for cancel icon

lex111:
Resolves #540


lex111:
@rwwagner90 done :heavy_check_mark:

test/unit/components/shepherd-footer.spec.js
```diff
@@ -58,10 +58,12 @@ describe('components/ShepherdFooter', () => {
     expect(cancelButton).toHaveAttribute('tabindex', '0');
     expect(cancelButton).toHaveClass('shepherd-button-secondary cancel-button shepherd-button');
     expect(cancelButton).toHaveTextContent('Exit');
+    expect(cancelButton).toHaveAttribute('aria-label', 'Exit');
```
rwwagner90: Could you please add a separate test for labels? We should test what happens when nothing is set for label, when label is set to null, undefined, string, number, etc and handle all the cases.

lex111: Is that good enough for you?

rwwagner90: @lex111 perhaps I was not clear enough. I would like a separate test for labels please. Not adding an `expect` to the existing test, but a separate test to check label functionality. It should cover all the different cases I mentioned: what happens when nothing is set for label, when label is set to null, undefined, string, number, etc

lex111: @rwwagner90 Sorry, I really misunderstood.
Need a separate `.spec` file for cancel button test? Or how best to do in this case?

rwwagner90: @lex111 no, just a separate test in the same file please

rwwagner90: @lex111 I can add the tests, if you would like

lex111: @rwwagner90 sorry for the long answer.. yes, please add tests, I’m unlikely to do it soon.

PR #643, opened 2019-10-25
Changed a condition

BSThisarasinghe:
I updated a condition


rwwagner90:
@BSThisarasinghe what was the reason for this change?

rwwagner90:
We don't need to switch this condition, so closing for now.

PR #644, opened 2019-10-28
Removed unnecessary parenthetical commas

sector-f:
I removed some commas that were syntactically placed as parenthetical commas despite the contained information not actually being parenthetical.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,7 @@ to contribute, we ask that you take the following steps:
 
 ## Testing
 
-All PRs, that change code functionality, are required to have accompanying tests.
+All PRs that change code functionality are required to have accompanying tests.
 
 ### Acceptance Tests
 


rwwagner90:
@sector-f I'm not sure I understand why we would want to remove these.

sector-f:
@rwwagner90 Parenthetical commas (e.g. the commas in question here) indicate that the information contained between them is not actually necessary for the meaning of the sentence. They're similar to parentheses--hence their name.

If the commas _should_ actually be there, then that means that the sentence really means "All PRs are required to have accompanying tests" (as opposed to just the PRs that change code functionality).

Removing the commas changes the sentence to mean that only PRs that change code functionality require accompanying tests. I have to assume that this is the correct interpretation of the sentence, since it really doesn't make sense for e.g. a PR that just fixes typos (like this one!) to have accompanying tests.

sector-f:
Also, it's October and I want my Hacktoberfest shirt.

rwwagner90:
@sector-f fair enough. We have some issues that need some help, if you would like to participate in more Hacktoberfest stuff.

PR #655, opened 2019-10-31
Reactively update title and buttons

VladimirAnaniev:
This change makes it so changes made with `updateStepOptions` actually update the template.  
The `$:` label syntax makes the statement reactive, and executes it on each update of the inputs

### DIFF
--- a/src/js/components/shepherd-button.svelte
+++ b/src/js/components/shepherd-button.svelte
@@ -3,8 +3,12 @@
   import { isFunction } from '../utils/type-check';
 
   export let config, step;
+  let action, classes, secondary, text, label;
 
-  $: ({ action, classes, secondary, text, label } = config);
+  $: {
+    action = config.action ? config.action.bind(step.tour) : null;
+    ({ classes, secondary, text, label } = config);
+  }
 
   let disabled = false;
 
@@ -57,7 +61,7 @@
   aria-label="{label ? label : null}"
   class="{`${(classes || '')} shepherd-button ${(secondary ? 'shepherd-button-secondary' : '')}`}"
   disabled={disabled}
-  on:click={action ? action.bind(step.tour) : null}
+  on:click={action}
   tabindex="0"
 >
     {text}
--- a/src/js/components/shepherd-button.svelte
+++ b/src/js/components/shepherd-button.svelte
@@ -3,7 +3,13 @@
   import { isFunction } from '../utils/type-check';
 
   export let config, step;
-  const { action, classes, secondary, text, label } = config;
+
+  $: action = config.action;
+  $: classes = config.classes;
+  $: secondary = config.secondary;
+  $: text = config.text;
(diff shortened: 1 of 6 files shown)


src/js/components/shepherd-button.svelte
```diff
@@ -3,7 +3,13 @@
   import { isFunction } from '../utils/type-check';
 
   export let config, step;
-  const { action, classes, secondary, text, label } = config;
+
+  $: action = config.action;
```
rwwagner90: @VladimirAnaniev would it be possible to make `config` itself reactive, rather than each individual property? It feels like there should be a less verbose way to accomplish this.

VladimirAnaniev: Config itself is reactive, but on line 6 we are getting the initial value and never updating it afterwards

VladimirAnaniev: I would love to make it more concise but could not find a better solution. The destructuring must be a part of variable declaration and that produced warnings with the label syntax.

Do you have an idea?

rwwagner90: Nice, glad that works!

rwwagner90:
@VladimirAnaniev I hate to ask, but would you perhaps be able to put some tests in showing these things update reactively as intended?

rwwagner90:
@VladimirAnaniev this looks good to me, but it looks like we have one test failure. It's something about `cy.click` being passed multiple elements. The fact that we didn't change anything in Cypress makes me concerned that these changes are creating multiple elements somehow.

VladimirAnaniev:
@rwwagner90 Seems like it was coming from the destructuring we discussed earlier. When I reverted it to the more verbose way as it was before, the tests passed. Will commit shortly

rwwagner90:
@VladimirAnaniev I don't understand. The destructuring seemed to work. I would love to figure out why the test was failing, and fix it, rather than abandoning destructuring.

VladimirAnaniev:
Sorry, not a svelte expert here, I could not figure out what exactly caused the issue. There was a problem with the title as well - when it was missing (first step in the demo) it was displayed as [object Object].  
I would also love to find the root cause. Do you have any ideas why it might be happening?  

This issue is stopping us from fully migrating to v6.

rwwagner90:
@VladimirAnaniev I will poke around a bit later today. If there isn't a good solution, we can go with this verbose one. Apologies for the delay.

VladimirAnaniev:
Thank you @rwwagner90, any updates on this?

rwwagner90:
@VladimirAnaniev we can merge this for now, and hopefully clean it up later.

VladimirAnaniev:
Thank you Robert, will it be possible to release a patch for this?

rwwagner90:
@VladimirAnaniev yeah, sure thing. I will kick that off now.

PR #665, opened 2019-11-07
Fix dynamic shepherd classes interfering with tether classes

VladimirAnaniev:
#655 Allowed the title to be changed reactively, but this caused an issue with the dynamically updated classes on the shepherd-element.  
The issue was that they overrode the classes tether is putting for position and the element disappeared altogether.  
By using the svelte class directive this will no longer be an issue as only the needed classes will be updated, not the whole class list.

### DIFF
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -165,7 +165,9 @@
   aria-describedby={!isUndefined(step.options.text) ? descriptionId : null}
   aria-labelledby={step.options.title ? labelId : null}
   bind:this={element}
-  class="{`${classes} shepherd-element ${hasCancelIcon ? 'shepherd-has-cancel-icon' : ''} ${hasTitle ? 'shepherd-has-title' : ''}`}"
+  class:shepherd-has-cancel-icon="{hasCancelIcon}"
+  class:shepherd-has-title="{hasTitle}"
+  class="{`${classes} shepherd-element`}"
   {...dataStepId}
   on:keydown={handleKeyDown}
   role="dialog"
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -1,20 +1,19 @@
 <script>
-  import { onMount } from 'svelte';
+  import { onMount, afterUpdate } from 'svelte';
   import ShepherdContent from './shepherd-content.svelte';
-  import { isUndefined } from '../utils/type-check.js';
+  import { isUndefined, isString } from '../utils/type-check.js';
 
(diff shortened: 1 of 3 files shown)


rwwagner90:
@VladimirAnaniev can you explain the need for this more? Was it that the `has-title` class was not applied if you dynamically added a title? Can we add some checks for some more of this dynamic stuff updating in some tests?

VladimirAnaniev:
The case was that when one of those fields changed (either removing/adding a title or enabling/disabling the cancelIcon), the class expression was re-evaluated and updated. This is fine, but what happened in result is that the tether related classes (`shepherd-attached-top`, etc.) were removed because svelte updated the class of the element to be what it evaluated, and it did not include the external classes.

rwwagner90:
@VladimirAnaniev ah, I see. We definitely do need to make sure we do not overwrite tether classes. This would still be a problem if you changed `classes` right?

VladimirAnaniev:
Turns out you cannot override `classes` as they are passed to the component on construction in `step.js` `_createTooltipContent` and the updateStepOptions method cannot change this

rwwagner90:
@VladimirAnaniev shouldn't you be able to update it? I thought you were able to update all step options?

VladimirAnaniev:
Not with the current way `classes` are handled in **shepherd-element.svelte**. It would need to be reworked to use `step.classes` instead of the property directly passed to the shepherd-element component in order to be ubdateable

rwwagner90:
Ah, I see. I think we should definitely update that. People will expect `updateStepOptions` to update any and all step options, I think.

VladimirAnaniev:
Sure. Should it be updated in this change? How do we handle the problem with the tether classes being overwritten?

rwwagner90:
I am not sure. @chuckcarpenter any ideas?

VladimirAnaniev:
Hate to ask but will it be possible to have this fix available in the meantime until a solution for the classes is available?

rwwagner90:
@VladimirAnaniev perhaps we should use `classList` manually for `classes`? I think we need to take custom classes out of the svelte class bindings to keep it from blowing away the tether classes.

VladimirAnaniev:
@rwwagner90 I implemented what you suggested and covered it with some test cases. Not the prettiest solution, but it works :)

VladimirAnaniev:
Thanks! I would appreciate it if this was published so we can finish the v6 migration :)

rwwagner90:
@VladimirAnaniev just pushed the 6.0.2 tag. Travis should build and deploy to npm shortly.

PR #752, opened 2020-01-08
[WIP] Popper v2

rwwagner90:
Giving Popper v2 a try, in the hopes that it is smaller and performs better than Tether.

- [x] Convert Modifiers to new format
- [x] Replace tests with Popper centric ones
- [x] Update docs to account for v2 syntax
- [x] Fix eslint in rollup


src/js/step.js
```diff
@@ -348,7 +348,8 @@ export class Step extends Evented {
 
     this.el.hidden = false;
 
-    this.tooltip.position();
+    // TODO figure out what the Popper v2 method is to force position recalculation
```
FezVrasta: It’s `instance.update()`, note that it returns a promise, so you can await for it if you need to

src/js/utils/general.js
```diff
@@ -85,65 +69,37 @@ export function uuid() {
...
-      ...tetherOptions.optimizations,
-      ...step.options.tetherOptions.optimizations
-    };
+  if (step.options.popperOptions) {
+    // TODO add popper modifiers support etc
 
-    tetherOptions = {
-      ...tetherOptions,
-      ...step.options.tetherOptions
+    popperOptions = {
```
FezVrasta: You may need a smarter merge algorithm here if you want to allow modifiers options to be merged with your default

atomiks:
The eslint problem is a bug with the plugin: https://github.com/TrySound/rollup-plugin-eslint/issues/40

Downgrade to 6.0.0 and it will work apparently

rwwagner90:
@atomiks downgrading throws other errors. We can keep it commented out for now.

src/js/components/shepherd-element.svelte
```diff
@@ -131,62 +131,42 @@
...
 
-  /**
-   * Arrow on bottom of tooltip centered horizontally
-   */
-  .shepherd-element.shepherd-element-attached-bottom.shepherd-element-attached-center .shepherd-arrow {
-    bottom: 0;
-    border-top-color: #fff;
-    left: 50%;
-    transform: translate(-50%, 100%);
+  .shepherd-arrow::before {
```
rwwagner90:
```suggestion
  .shepherd-arrow:before {
```

src/js/utils/general.js
```diff
@@ -85,65 +71,42 @@ export function uuid() {
   return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
     const r = (d + Math.random() * 16) % 16 | 0;
     d = Math.floor(d / 16);
-    return (c == 'x' ? r : (r & 0x3 | 0x8)).toString(16);
+    return (c == 'x' ? r : (r & 0x3) | 0x8).toString(16);
```
rwwagner90: Was this intentional?

test/cypress/integration/a11y.spec.js
```diff
@@ -62,12 +62,12 @@ describe('a11y', () => {
       cy.get('.shepherd-element').should('not.exist');
     });
 
-    it.skip('Tab is focus trapped inside the modal', () => {
+    it('Tab is focus trapped inside the modal', () => {
```
rwwagner90: 🔥

test/unit/utils/general.spec.js
```diff
@@ -12,50 +12,4 @@ describe('General Utils', function() {
...
-            element: 'bar'
-          }
-        }
-      });
-
-      const tetherOptions = getTetherOptions(parseAttachTo(step), step);
-      expect(tetherOptions.classes.element).toBe('bar');
-    });
-
-    it('constraints can be overridden', function() {
```
rwwagner90: Should we replace these tests with some `popperOptions` tests? Would be good to make sure deeply nested options can be passed and override ours. I had issues with this in the past.

docs-src/tutorials/02-usage.md
```diff
@@ -265,9 +268,9 @@ yourApp.on('some-event', () => {
 
 ### Rendering Tours in Specific Locations
 
-By default, tour steps will append their elements to the `body` element of the DOM. This is perfect for most use cases, but not always. If you need to have steps appended elsewhere you can take advantage of Tippy's
-[`appendTo` option](https://atomiks.github.io/tippyjs/#append-to-option) by defining it on the
-`tippyOptions` hash inside of each Step's options hash.
+By default, tour steps will append their elements to the `body` element of the DOM. This is perfect for most use cases, but not always. If you need to have steps appended elsewhere you can take advantage of Popper's
+[options](https://popper.js.org/docs/v2/constructors/#options) by defining it on the
```
rwwagner90: What popper option would be used for this? We may want to be explicit here.

test/unit/utils/general.spec.js
chuckcarpenter: I was a little back and forth on this, but should it be on the implementor to understand what they're passing in and use v2 docs?

docs-src/tutorials/02-usage.md
chuckcarpenter: Actually, this is a tippy option and we easily do this with `Step.target`. Going to remove.

patches/rollup-plugin-eslint+7.0.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/rollup-plugin-eslint/index.js b/node_modules/rollup-plugin-eslint/index.js
```
rwwagner90: Woah, fancy

test/unit/utils/general.spec.js
rwwagner90: We should add tests for the options merging is what I mean. like `options.modifiers.foo.bar` and make sure that overrides our defaults and can still be passed down, and we don't clobber it. Does that make sense?

rwwagner90: I had a lot of issues before with using `Object.assign` and accidentally clobbering nested options.

docs-src/tutorials/02-usage.md
rwwagner90: I don't think `step.target` is the same thing. This is defining the container the step element itself lives in right?

chuckcarpenter: if you look at the docs on `createPopper()` The first arg is target and we map target there. Maybe I'm not clear on what you mean then?

rwwagner90: Target is what you are attaching the popper to. `appendTo` is where the tooltip element itself lives.

rwwagner90:
@chuckcarpenter I think the only thing missing from this PR is constraining to window. i.e. when the window shrinks and there is no room to flip to the other side, the tooltips should start moving to the middle of the target, to allow them to be read.

PR #797, opened 2020-02-13
🐛 Fix step positioning for viewport

chuckcarpenter:


chuckcarpenter:
@rwwagner90 running `build:landing` add a bunch, so let me know if that looks right.

PR #801, opened 2020-02-17
🐛 Fix initial scroll to top before scrolling to target

chuckcarpenter:
This fixes the issue with the document scrolling to top before each step is shown. Also, removes extra renders via the forced update in `_step()` method.

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -7,7 +7,6 @@ import ShepherdElement from './components/shepherd-element.svelte';
 
 // Polyfills
 import smoothscroll from 'smoothscroll-polyfill';
-
 smoothscroll.polyfill();
 
 /**
@@ -329,7 +328,6 @@ export class Step extends Evented {
     if (this.options.advanceOn) {
       bindAdvance(this);
     }
-
     setupTooltip(this);
   }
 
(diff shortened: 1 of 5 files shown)


src/js/step.js
```diff
@@ -345,22 +343,22 @@ export class Step extends Evented {
 
     this.tour.modal.setupForStep(this);
     this._styleTargetElementForStep(this);
-
     this.el.hidden = false;
 
-    this.tooltip.update();
```
rwwagner90: Are you sure we can remove this? Perhaps this was the wrong approach, but I think we needed to call `update` to ensure things got repositioned when scrolling or resizing the page etc.

chuckcarpenter: this seems to do extra renders, does the eventlistener for scroll call the initial show function?

rwwagner90: Unsure, but we need an update somewhere I think. Maybe I am wrong and popper now handles this internally?

rwwagner90: Does it update on scroll and resize without this?

chuckcarpenter: when it is out of viewport, there's a data attr added and we use that to set opacity to 0. What behavior where you expecting. Also, resize seems to have it's own internal events, as it does work as I'd expect.

test/unit/utils/general.spec.js
```diff
@@ -29,7 +29,7 @@ describe('General Utils', function() {
         }
       });
 
-      const { popperOptions } = getPopperOptions(parseAttachTo(step), step);
```
chuckcarpenter: scope the test to just one method

src/js/step.js
rwwagner90: Just ensuring that the popper moves around and tries to stay in the viewport. I think on resize and scroll, the popper would stay in one spot before, if we weren't calling `update`.

rwwagner90: I checked this out locally, and it seems to be fine 👍

PR #802, opened 2020-02-18
🔥 Remove demo directory and use landing site as default

chuckcarpenter:
This removes the extra directory that was the old demo site and uses landing as both testing and local dev site.


rollup.config.js
```diff
@@ -54,15 +54,14 @@ if (process.env.DEVELOPMENT) {
...
           '/dist/js/shepherd.js': 'dist/js/shepherd.js',
-          '/demo/js/prism.js': 'demo/js/prism.js',
-          '/demo/js/welcome.js': 'demo/js/welcome.js',
-          '/demo/css/prism.css': 'demo/css/prism.css',
-          '/demo/css/welcome.css': 'demo/css/welcome.css',
-          '/demo/sheep.svg': 'demo/sheep.svg'
+          '/landing/js/prism.js': 'landing/js/prism.js',
+          '/landing/js/welcome.js': 'landing/js/welcome.js',
+          '/landing/css/prism.css': 'landing/css/prism.css',
+          '/landing/css/welcome.css': 'landing/css/welcome.css',
```
rwwagner90: I'm not sure that just changing `demo` to `landing` is going to do what you expect. Doesn't the landing page need a tailwind build?

chuckcarpenter: I guess I just don't understand where `build:landing` is ever run? I don't see it in CI.

chuckcarpenter: added a postbuild script

rwwagner90: `build:landing` is not run, I don't think. It has to manually be run to add tailwind styles. I think what I am getting at is live reload isn't going to work without tailwind build stuff built in. We probably need a tailwind rollup plugin or something here.

PR #811, opened 2020-02-25
Fix bug when modifiers was not defined

rwwagner90:


src/js/utils/general.js
```diff
@@ -106,9 +109,14 @@ export function getPopperOptions(attachToOptions, step) {
   }
 
   if (step.options.popperOptions) {
-    if (step.options.popperOptions.modifiers.length > 0) {
+    if (
+      step.options.popperOptions.modifiers &&
```
rwwagner90: This was the main change here

PR #824, opened 2020-03-04
Single argument on classList.add for IE11 compatibility

rikvdlooi:
According to MDN web docs (https://developer.mozilla.org/en-US/docs/Web/API/Element/classList and https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/add) Internet Explorer doesn't support multiple arguments for classList.add().
This adds the two classes is separate statements to keep compatibility with IE11.


rwwagner90:
Thanks for the PR!

PR #825, opened 2020-03-05
🐛 Fixes modifier override order for defaultStepOptions

### LINKED ISSUE #816: defaultStepOptions.popperOptions only seems to apply to first step
Given the following input:
```js
    tourName: "Demo",
    // ...
    defaultStepOptions: {
      scrollTo: { behavior: "smooth", block: "center" },
      popperOptions: {
        modifiers: [{ name: "offset", options: { offset: [0, 32] } }]
      },
    }
  });

 // ...
 tour.addSteps([
   // ...
 ]);
```

`defaultStepOptions.popperOptions` (i.e. the popper offset) only seems to apply to the first step in the tour. The other default option, `scrollTo`, applies to each step just fine.

Any insights?

chuckcarpenter:
closes #816

### DIFF
--- a/landing/js/welcome.js
+++ b/landing/js/welcome.js
@@ -18,9 +18,6 @@
         scrollTo: {
           behavior: 'smooth',
           block: 'center'
-        },
-        popperOptions: {
-          modifiers: [{ name: "offset", options: { offset: [0, 32] } }]
         }
       },
       // This should add the first tour step
@@ -46,17 +43,7 @@
               text: 'Next'
             }
           ],
-          id: 'welcome',
-          popperOptions: {
-            modifiers: [
-              {
-                name: 'preventOverflow',
-                options: {
-                  altAxis: false
-                }
-              }
-            ]
-          }
+          id: 'welcome'
         }
       ],
       useModalOverlay: true
--- a/landing/js/welcome.js
+++ b/landing/js/welcome.js
@@ -46,7 +46,17 @@
               text: 'Next'
             }
           ],
-          id: 'welcome'
+          id: 'welcome',
+          popperOptions: {
+            modifiers: [
+              {
+                name: 'preventOverflow',
+                options: {
+                  altAxis: false
+                }
+              }
+            ]
+          }
         }
       ],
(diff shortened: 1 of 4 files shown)


src/js/step.js
```diff
@@ -289,6 +289,7 @@ export class Step extends Evented {
    * @private
    */
   _setOptions(options = {}) {
+
```
rwwagner90:
```suggestion
```

src/js/utils/general.js
```diff
@@ -118,15 +118,10 @@ export function getPopperOptions(attachToOptions, step) {
         (mod) => !names.includes(mod.name)
       );
 
-      popperOptions.modifiers = [
-        ...step.options.popperOptions.modifiers,
-        ...filteredModifiers
-      ];
-
-      delete step.options.popperOptions.modifiers;
+      popperOptions.modifiers = Array.from(new Set([...popperOptions.modifiers, ...filteredModifiers, ...step.options.popperOptions.modifiers]));
```
rwwagner90: What's the browser support on `Array.from`?

landing/js/welcome.js
```diff
@@ -18,6 +18,9 @@
         scrollTo: {
           behavior: 'smooth',
           block: 'center'
+        },
+        popperOptions: {
+          modifiers: [{ name: "offset", options: { offset: [0, 32] } }]
```
rwwagner90: Do we want this offset on the landing or was it leftover from where you were playing with it?

chuckcarpenter: I think it's nice to have these examples in while developing, but I can remove if you prefer.

rwwagner90: Does this not change the style of the landing page though? I think we would want to keep the landing page styled as it was, right?

src/js/utils/general.js
chuckcarpenter: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility

PR #851, opened 2020-03-17
➖ Remove references to IE11

chuckcarpenter:


sidzaveri:
@rwwagner90 @chuckcarpenter does this mean that IE 11 compatibility has been officially dropped for all future versions as well?

PR #852, opened 2020-03-18
🚧 Test different action for full dir publishing

chuckcarpenter:


rwwagner90:
@chuckcarpenter where are we on this?

PR #866, opened 2020-03-27
Update shepherd-button.svelte to allow HTML

ScottHuston:
Make shepherd-button capable of displaying HTML.  Without @html Svelte converts the value into a textNode, which escapes HTML. 

#865


ScottHuston:
Yes it does.

PR #892, opened 2020-04-07
fix: add missing type definition

chanlito:


src/types/step.d.ts
```diff
@@ -187,6 +187,11 @@ declare namespace Step {
      * ```
      */
     when?: StepOptionsWhen;
+
+    /**
+     * A boolean, that when set to false, will set `pointer-events: none` on the target.
+     */
+    canClickTarget?: boolean;
```
rwwagner90, line 194: This should be added to StepOptions, like `when` is above. Also, can we please ensure this is in alphabetical order?

chanlito: Adjusted.

PR #894, opened 2020-04-09
🐛 Add new function for creating modals when needed and cleanup

chuckcarpenter:
This fixes the repeat overlay issue reported in https://github.com/shipshapecode/react-shepherd/issues/288


test/unit/components/shepherd-modal.spec.js
```diff
@@ -1,5 +1,5 @@
 import ShepherdModal from '../../../src/js/components/shepherd-modal.svelte';
-import { Tour } from '../../../src/js/tour.js';
+// import { Tour } from '../../../src/js/tour.js';
```
rwwagner90, line 2: If we don't need Tour anymore, can we delete this import?

package.json
```diff
@@ -43,7 +43,7 @@
     "test:cy:ci:firefox": "start-server-and-test start-test-server http://localhost:9002 cy:run:firefox",
     "test:cy:watch": "start-server-and-test start-test-server http://localhost:9002 cy:open",
     "test:unit:ci": "jest --coverage",
-    "test:unit:watch": "jest --watch",
+    "test:unit:watch": "jest --watch -t 'shepherdModalOverlayContainer'",
```
rwwagner90, line 46: What is this ` -t 'shepherdModalOverlayContainer'` for?

chuckcarpenter: text filter for running just suites or single tests.

PR #908, opened 2020-04-14
🐛 Fix centered sizing and add max height defaults for text

chuckcarpenter:
closes: #837 

mobile:
image

and desktop:
image

### DIFF
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -120,7 +120,7 @@
     opacity: 0;
     outline: none;
     transition: opacity 0.3s;
-    width: 400px;
+    width: 100%;
     z-index: 9999;
   }
 
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -120,6 +120,7 @@
     opacity: 0;
     outline: none;
     transition: opacity 0.3s;
+    width: 400px;
     z-index: 9999;
   }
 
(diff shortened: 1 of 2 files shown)


src/js/components/shepherd-element.svelte
```diff
@@ -120,6 +120,7 @@
     opacity: 0;
     outline: none;
     transition: opacity 0.3s;
+    width: 400px;
```
chuckcarpenter, line 123: This fixes this issue in mobile
image

rwwagner90: Hmm, I'm not sure if we want both `width` and `max-width`. Perhaps it should be `width:100%` with `max-width: 400px`?

src/js/components/shepherd-text.svelte
```diff
@@ -24,6 +24,8 @@
     color: rgba(0, 0, 0, 0.75);
     font-size: 1rem;
     line-height: 1.3em;
+    max-height: 30vh;
```
rwwagner90, line 27: Should we make this different for desktop/mobile or do you think this adequately works for most cases? I think I might prefer a max height of something like `75%`.

chuckcarpenter: it kind of looks weird on desktop, with the max width of 400px, but I guess that's subjective.

chuckcarpenter: also, it covers the element being pointed to, e.g.,

image

PR #911, opened 2020-04-15
Corrections in the docs

### LINKED ISSUE #910: Errors in the docs
Found several errors in the docs:

## Step

There are still several mentions of `tippy` in the docs.

---

arrow | Object | Whether to display the arrow for the tooltip or not.
-- | -- | --

- `Object` should be `Boolean`
- Maybe add the defaults (`true` in this case)?

---

```js
const new Step(tour, {
  attachTo: { element: '.some .selector-path', on: 'left' },
  ...moreOptions
})'
```

- First line should start with `const step = new Step(...`
- Replace trailing `'` with `;` on last line
- Same goes for the other examples

sn3p:
closes #910

### DIFF
--- a/docs-src/tutorials/02-usage.md
+++ b/docs-src/tutorials/02-usage.md
@@ -4,7 +4,7 @@ The latest versions of Rollup and Webpack support ES6 imports. We have an ES mod
 exported to `dist/js/shepherd.esm.js`. This is also specified as `"module"` in
 `package.json`, which should allow you to import using standard ES import syntax.
 
-i.e. 
+i.e.
 
 ```js
 import Shepherd from 'shepherd.js';
@@ -30,8 +30,8 @@ Next, add your steps:
 tour.addStep({
   id: 'example-step',
   text: 'This step is attached to the bottom of the <code>.example-css-selector</code> element.',
-  attachTo: { 
-    element: '.example-css-selector', 
+  attachTo: {
+    element: '.example-css-selector',
     on: 'bottom'
   },
   classes: 'example-step-extra-class',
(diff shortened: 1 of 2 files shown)


docs-src/tutorials/02-usage.md
```diff
@@ -157,15 +157,15 @@ created.
 - `title`: The step's title. It becomes an `h3` at the top of the step.
 - `attachTo`: What element the step should be attached to on the page.
 It should be an object with the properties `element` and `on`, where `element` is an element selector string
-or a DOM element and `on` is the optional direction to place the Tippy tooltip.
-              
+or a DOM element and `on` is the optional direction to place the step element.
```
rwwagner90, line 160:
```suggestion
or a DOM element and `on` is the optional direction to place the Popper tooltip.
```

src/js/step.js
```diff
@@ -133,7 +133,7 @@ export class Step extends Evented {
   }
 
   /**
-   * Remove the step, delete the step's element, and destroy the tippy instance for the step
+   * Remove the step and delete the step's element
```
rwwagner90, line 136:
```suggestion
   * Remove the step, delete the step's element, and destroy the Popper instance for the step.
```

docs-src/tutorials/02-usage.md
sn3p: Popper is not really a "tooltip", but can do 👍

rwwagner90: We're creating tooltips with Popper though

rwwagner90:
Thanks for the PR!

docs-src/tutorials/02-usage.md
sn3p: True :)

PR #919, opened 2020-04-21
🐛 Create deep copy of tour options to make immutable

### LINKED ISSUE #915: Layout error when has popperOptions
Reproduce link: https://jsfiddle.net/br8o467m/12/
If have `popperOptions`, the step what has `attachTo` will error
```js
  defaultStepOptions: {
    ...
    popperOptions: {}
    ...
  },
});
```

chuckcarpenter:
closes #915

The shallow copy created via `Object.assign` was causing the original tour `defaultStepOptions` to be mutated by step options. More info: 

Object.assign({})
Can only make shallow copies of objects so it will only work in a single level (first level) of the object reference. 

Object spread:
Object spread does a 'shallow copy' of the object. Only the object itself is cloned, while "nested instances are not cloned". This results in mutated nested objects.

JSON.parse(JSON.stringify()):
This is going to work fine as long as your Objects and the nested Objects "only contains primitives". This is going to fail when it is passing functions on values such as `beforeShowPromise`.

@rwwagner90 otherwise, we're going to need to use a much more complex deep merge utility, whether written ourselves or something like https://github.com/kolodny/immutability-helper

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -297,7 +297,8 @@ export class Step extends Evented {
       {
         arrow: true
       },
-      tourOptions,
+      // make a deep copy in order to prevent mutation on the source
+      JSON.parse(JSON.stringify(tourOptions)),
       options
     );
 
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -292,13 +292,13 @@ export class Step extends Evented {
   _setOptions(options = {}) {
     const tourOptions =
       this.tour && this.tour.options && this.tour.options.defaultStepOptions;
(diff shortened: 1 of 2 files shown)


rwwagner90:
@chuckcarpenter have you taken a look at how we merged options in Popper v1? There was some logic that handled these cases, I believe.

Basically:

```js
popperOptions = {
...defaultStepOptions.popperOptions,
...step.popperOptions
}

popperOptions.modifiers = {
...defaultStepOptions.popperOptions.modifiers,
...step.popperOptions.modifiers
}
```

Obviously, this example doesn't totally work, but just illustrating the idea to use spread at various levels to achieve a somewhat deeper merge of certain properties.

PR #921, opened 2020-04-23
Deep copy defaultStepOptions

### LINKED ISSUE #915: Layout error when has popperOptions

rwwagner90:
This avoids keeping a polluted state around when setting up multiple step options.

Fixes #915

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -290,8 +290,10 @@ export class Step extends Evented {
    * @private
    */
   _setOptions(options = {}) {
-    const tourOptions =
+    let tourOptions =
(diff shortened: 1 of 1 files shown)


src/js/step.js
```diff
@@ -290,9 +290,11 @@ export class Step extends Evented {
    * @private
    */
   _setOptions(options = {}) {
-    const tourOptions =
```
chuckcarpenter, line 293: use `const` when no reassigning

PR #935, opened 2020-04-29
cookbook - progress bar code snippet fix  to show in all steps

Hariharan-Ingersal:
Since `document.querySlector` is used, the `document.querySelector('.shepherd-header')` would still give the header of step 1 even if it is executed for step 2. Hence changed it to use step element 'el' instead of document.


rwwagner90:
Thanks for the PR!

PR #946, opened 2020-05-07
Fix content being blocked by hidden steps

### LINKED ISSUE #934: Content is blocked by "hidden" steps
When a tour is hidden because the target is not visible on the page (e.g. because of a media query), it is blocking/masking the content behind it. This is because it's still rendered (always in the top-left corner of the document) with just `opacity: 0;`.

Is this a known problem?

### Demo
https://codepen.io/snap/pen/yLYXQmG

### Possible fix
```css
.shepherd-element[data-popper-reference-hidden]:not(.shepherd-centered) {
  opacity: 0;
  pointer-events: none; /* <- added */
}
```

rwwagner90:
Fixes #934

### DIFF
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -119,7 +119,7 @@
     max-width: 400px;
     opacity: 0;
     outline: none;
-    transition: opacity 0.3s;
+    transition: opacity 0.3s, visibility 0.3s;
     visibility: hidden;
     width: 100%;
     z-index: 9999;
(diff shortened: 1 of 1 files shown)


src/js/components/shepherd-element.svelte
```diff
@@ -120,16 +120,19 @@
     opacity: 0;
     outline: none;
     transition: opacity 0.3s;
```
sn3p, line 122: You probably want to add `visibility` to the transitioned properties:

```suggestion
    transition: opacity 0.3s, visibility 0.3s;
```

Otherwise it will instantly show the step and you won't see the `opacity` transition.
This feels a bit counter intuitive as `visibility` isn't a calculable property, but it works :)

https://codepen.io/snap/pen/dyYdjQe

PR #947, opened 2020-05-07
Bump deps, yarn upgrade, fix tests

rwwagner90:


src/js/utils/general.js
```diff
@@ -99,12 +99,19 @@ export function getPopperOptions(attachToOptions, step) {
...
           altAxis: true
         }
+      },
+      {
+        name: 'focusAfterRender',
+        enabled: true,
+        phase: 'afterWrite',
+        fn() {
+          setTimeout(() => {
+            step.el.focus();
```
rwwagner90, line 109: @chuckcarpenter this modifier is required if we want to show/hide the popper when scrolling away. We have to focus every time the popper recalculates basically, so if the element is shown it gets focused.

test/cypress/integration/a11y.spec.js
```diff
@@ -38,6 +38,14 @@ describe('a11y', () => {
 
     beforeEach(() => {
       tour = setupTour(Shepherd);
+
+      // This is a hack removing the extra page elements so the page does not scroll. Cypress hates scrolling for some reason.
```
rwwagner90, line 42: @chuckcarpenter this hack sucks, but Cypress was refusing to scroll right, and since the scrolling was impossible to get to work, the poppers were always hidden, thus breaking the tests.

chuckcarpenter: reasonable, but I guess we'll need to figure out a way to test in scrolling at some point.

PR #959, opened 2020-05-12
Add deepmerge for deep clone functionality

### LINKED ISSUE #945: defaultStepOptions showOn and when option no longer work after update to 7.x
defaultStepOptions showOn and when option no longer work after update to 7.x, it does work when adding the options directly to the step.

```
        this.tour = new Shepherd.Tour({
            tourName: CURRENT_TOUR_NAME,
            useModalOverlay: true,
            defaultStepOptions: {
                classes: 'shepherd-theme',
                canClickTarget: false,
                showOn: function () {
                    return hasNotSeenCurrentStep(this.id);
                },
                when: {
                    show: function () {
                        addSeenTourStep(this.id);
                        trackUserSeenStep(this.id);
                    },
                },
                cancelIcon: {
                    enabled: true,
                    label: 'Stoppen met de uitleg',
                },
            },
        });
```

Is something changed in the way I provide these settings ?

rwwagner90:
Fixes #945

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -1,3 +1,4 @@
+import merge from 'deepmerge';
 import { Evented } from './evented.js';
 import autoBind from './utils/auto-bind.js';
 import { isElement, isFunction, isUndefined } from './utils/type-check.js';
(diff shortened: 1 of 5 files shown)


src/js/step.js
```diff
@@ -309,7 +310,7 @@ export class Step extends Evented {
     let tourOptions =
       this.tour && this.tour.options && this.tour.options.defaultStepOptions;
 
-    tourOptions = tourOptions ? JSON.parse(JSON.stringify(tourOptions)) : {};
+    tourOptions = merge({}, tourOptions || {});
```
rwwagner90, line 313: @chuckcarpenter this is essentially the only logic change, switching `JSON.stringify` to use `deepmerge` instead so we get a deep clone.

PR #982, opened 2020-05-24
Officially drop IE 11 and remove from browserslist

rwwagner90:


sn3p:
What will break in IE11, and what is needed to make it work in IE11?
We still need support for IE11, but I hope not for long 😭

rwwagner90:
@sn3p we already were not really supporting IE11, so if it was working for you before, I would imagine it will still work.

PR #995, opened 2020-06-03
Added the correct this context to StepOption 'when' functions

WORMSS:
Update step.d.ts
With accordance to https://github.com/shipshapecode/shepherd/blob/ad5b9b0daa3a2db1a507b70f6f4d6b5620add40d/src/js/step.js#L332 the "this" context is set to the Step

### DIFF
--- a/src/types/step.d.ts
+++ b/src/types/step.d.ts
@@ -279,7 +279,7 @@ declare namespace Step {
   }
 
   interface StepOptionsWhen {
-    [key: string]: (() => void);
+    [key: string]: ((this: Step) => void);
   }
 }
 


rwwagner90:
@WORMSS I do not know TypeScript very well, but should we also indicate that the `this` is the step here?

https://github.com/shipshapecode/shepherd/blob/99351de7073540334660c3f840b086d8dfd60f68/src/types/step.d.ts#L234

WORMSS:
Ahh, I was unaware of actions. As the code was not in the JS but in the svelte and I had never used svelte before.
If I am reading this correctly, would it be the tour rather than Step?
https://github.com/shipshapecode/shepherd/blob/bec705e79b45f0fc942b759ca4ece6874f162d90/src/js/components/shepherd-button.svelte#L8
Which would make the Typescript
```diff
- action?: (() => void);
+ action?: ((this: Tour) => void);
```

rwwagner90:
@WORMSS yeah, that is correct.

PR #1026, opened 2020-07-02
Add Edge 18 to browserslist

### LINKED ISSUE #1023: Doesn't work in Microsoft Edge (EdgeHTML 18)
Open https://shepherdjs.dev/ in Microsoft Edge and you can't access the demo because of a syntax error in the script. 
`SCRIPT1028: SCRIPT1028: Expected identifier, string or number`

The problem is the `...popperOptions` object spread in the code below which is not supported in the (old) EdgeHTML 18 version.
https://caniuse.com/#feat=mdn-javascript_operators_spread_spread_in_object_literals

Your browserlist specifies `last 2 Edge versions` which apparently doesn't include Edge 18 anymore (see Can I Use) but according to https://browserl.ist/?q=last+2+Edge+versions it still does. I'm not sure what to make of this.

Transpilation should fix the problem, so including `Edge 18` in your browserlist should be enough. Alternatively maybe the documentation should explicitly state support for "Chromium Edge" only. The problem with that is that according to Can I Use EdgeHTML has a global usage of 2% and Chromium Edge is still at 0%.

```js
export function makeCenteredPopper(step) {
  const centeredStylePopperModifier = _getCenteredStylePopperModifier();

  let popperOptions = {
    placement: 'top',
    strategy: 'fixed',
    modifiers: [
      {
        name: 'focusAfterRender',
        enabled: true,
        phase: 'afterWrite',
        fn() {
          setTimeout(() => {
            if (step.el) {
              step.el.focus();
            }
          }, 300);
        }
      }
    ]
  };

  popperOptions = {
    ...popperOptions,
    modifiers: Array.from(
      new Set([...popperOptions.modifiers, ...centeredStylePopperModifier])
    )
  };

  return popperOptions;
}
```

te1:
- Enables transpilation of object property spread
- Required for (old) EdgeHTML 18 compatibility
- Fixes #1023

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -4,3 +4,4 @@ last 2 Chrome versions
 last 2 Firefox versions
 last 2 Safari versions
 last 2 Edge versions
+Edge 18


rwwagner90:
Thanks for the PR!

PR #1028, opened 2020-07-05
_scrollTo works with SVGElement too

### LINKED ISSUE #1027: scrollTo a SVGElement wont scroll
Hello,

For the scroll behavior, you are using https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView by default. It is implemented on Element but the `isElement` function of type-checks compares against HTMLElement and not Element or SVGElement :
https://github.com/shipshapecode/shepherd/blob/356e3859df550bb55cf6a25c9be36a1b6f4ca3e2/src/js/step.js#L289-L294

Thus, scrolling to a SVG element won't enter the else, and wont scroll at all.

I recommend to use `typeof Element` condition.

BTW this causes a bug on your demo page (https://shepherdjs.dev/), when you hit NEXT then BACK.

Hypher:
Fixes #1027 using instanceof Element to target both HTMLElement and SVGElement

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -287,7 +287,7 @@ export class Step extends Evented {
     if (isFunction(this.options.scrollToHandler)) {
       this.options.scrollToHandler(element);
     } else if (
-      isElement(element) &&
+      element instanceof Element &&
       typeof element.scrollIntoView === 'function'
     ) {
       element.scrollIntoView(scrollToOptions);


src/js/step.js
```diff
@@ -287,7 +287,7 @@ export class Step extends Evented {
     if (isFunction(this.options.scrollToHandler)) {
       this.options.scrollToHandler(element);
     } else if (
-      isElement(element) &&
+      element instanceof Element &&
```
rwwagner90, line 290: Perhaps we should change `isElement` to check for `Element` instead?

Hypher: `isElement `is used elsewhere and is not documented this way. It checks only `HTMLElement`.

Especially, it will create a bug in: https://github.com/shipshapecode/shepherd/blob/4885a3e7c9668005d71d0de70296f553b6063935/src/js/tour.js#L298-L299

Because `focus` is a method of `HTMLElement`, not `Element`.

If you really want to implement it in type-check.js, then I suggest :
 - Either rename `isElement` -> `isHTMLElement`; and implement an `isElement` that checks for `Element`
 - Or implement an `isAnyElement` that checks for `Element`

rwwagner90: Let's rename `isElement` to `isHTMLElement` and add an `isElement` that checks for `Element` please.

Hypher: I'm sorry, I won't have time to make this refactoring, would you like to do it?

PR #1044, opened 2020-07-13
add Step.id to typings, fix tour.getById() typing

cyremur:
Exposes the `id` property of `Step` in TypeScript.
Fix the typing of `Tour.getById()` which can return null if given an `id` that's not in the tour.


rwwagner90:
Thanks so much for the PR @cyremur!

PR #1075, opened 2020-07-31
Add container property to step element

nikitamadin:
Hello, it was necessary to place step elements inside the container, so I added the `elementContainer` property

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -260,7 +260,7 @@ export class Step extends Evented {
     const labelId = `${this.id}-label`;
 
     this.shepherdElementComponent = new ShepherdElement({
-      target: document.body,
+      target: this.tour.options.defaultStepOptions.elementContainer || document.body,
       props: {
         classPrefix: this.classPrefix,
         descriptionId,


src/js/step.js
```diff
@@ -260,7 +260,7 @@ export class Step extends Evented {
     const labelId = `${this.id}-label`;
 
     this.shepherdElementComponent = new ShepherdElement({
-      target: document.body,
+      target: this.tour.options.defaultStepOptions.elementContainer || document.body,
```
rwwagner90, line 263: If this is a tour property, it should go on the tour. It will also need docs and tests.

shepmaster:
@nikitamadin do you believe you will have time to push this feature to completion? I'm looking for something similar.

nikitamadin:
@shepmaster no, sorry

src/js/step.js
shepmaster: > It will also need docs and tests.

@rwwagner90 can you give some intro guidance as to what you'd like to see here?

rwwagner90:
@shepmaster would you have the time to open a new PR finishing this? I'm a little unclear on the use case and whether this should be a tour or step option. If it should be a step option, we shouldn't use `defaultStepOptions` but instead `step.options` etc.

rwwagner90:
@shepmaster thoughts on continuing this?

rwwagner90:
Closed by #1176

PR #1119, opened 2020-08-25
shift + tab handled

### LINKED ISSUE #1076: shift + tab on init breaks out of tour modal
Replication steps:

1. Load tour
2. Use the keyboard to select SHIIFT + TAB

Expected behaviour; focus should be contained within the tour box.
Actual behaviour; focus breaks out of tour box and to the last element on the page.  

This behaviour is replaceable here https://shepherdjs.dev/

faizanu94:
This pull request closes #1076 by adding an additional check on `data-popper-placement` attribute on `document.activeElement` when `shiftKey` is pressed

### DIFF
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -79,7 +79,7 @@
         }
         // Backward tab
         if (e.shiftKey) {
-          if (document.activeElement === firstFocusableElement) {
+          if (document.activeElement === firstFocusableElement || document.activeElement.hasAttribute("data-popper-placement")) {
             e.preventDefault();
             lastFocusableElement.focus();
           }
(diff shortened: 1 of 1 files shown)


src/js/components/shepherd-element.svelte
```diff
@@ -79,7 +79,7 @@
         }
         // Backward tab
         if (e.shiftKey) {
-          if (document.activeElement === firstFocusableElement) {
+          if (document.activeElement === firstFocusableElement || document.activeElement.hasAttribute("data-popper-placement")) {
```
rwwagner90, line 82: Would it perhaps make more sense to check for a class of `shepherd-element`?

faizanu94: That would be perfect! Thanks for the suggestion

PR #1132, opened 2020-09-04
Added default fallback for target when step is hidden/destroyed

### LINKED ISSUE #1131: "shepherd-enabled" class is not removed from body element when tour ends
Reproducible on the Shepherd demo site: https://shepherdjs.dev/

Reach the end of the demo and inspect the body element - "shepherd-enabled" will still be set.

This is causing an issue in a project where we disable some scrollbars on certain elements if **shepherd-enabled** is set on the body. As it's never removed, the scrollbars remain disabled when the tour has been completed/cancelled.

IWMTom:
Fixes #1131 

If a step is added without specifically specifying an element to attach to, there was a default to fallback to **document.body**.
When a step was hidden/destroyed, **this.target** was checked which in the case of a non-attached step, **this.target** was null.

- Added a fallback to **document.body** if **this.target** is falsey when hiding/destroying a step
- Removed condition of **this.target** being truthy for calls to **_updateStepTargetOnHide()** 
- Added unit tests to cover this scenario. Validated that the test failed prior to the fix, and that the test now passes.

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -155,9 +155,7 @@ export class Step extends Evented {
       this.el = null;
     }
 
-    if (this.target) {
-      this._updateStepTargetOnHide();
-    }
+    this._updateStepTargetOnHide();
 
     this.trigger('destroy');
   }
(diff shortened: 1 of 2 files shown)


rwwagner90 (approved):

rwwagner90:
Thanks for the PR!

PR #1187, opened 2020-10-06
Implement lazy evaluation of `attachTo.element`

zefj:
This PR introduces lazy evaluation of `attachTo.element`,  and memoization of it's result. To do that, it extends `attachTo.element` signature to also accept a function. `attachTo.element` is resolved in the `before-show` phase. A by-product is that string selectors are now also resolved in the same phase, and only once (which is likely, but depends on usage and therefore is not guaranteed - see below for explanation).

New interface of `StepOptionsAttachTo`:

```js
  interface StepOptionsAttachTo {
    element?: HTMLElement | string | (() => HTMLElement | string | null | undefined);
    on?: PopperPlacement;
  }
```

## Benefits:

1. It is now simplier to tour dynamic content - the `attachTo` elements do not have to exist in DOM during tour creation, or even at tour start. The element to attach to is evaluated on demand, during the `before-show` phase, triggered by `Step._show()` method right before setting up step elements.
2. The library gets much more robust - we memoize the resolver result and use that for all logic required to render a step. Previously, if you supplied a selector string to `attachTo.element`, the library would call `document.querySelector` on it approximately 4 times for every `_show`. That's 4 places for potential failures.
3. Step options can now be immutable - previously, if you wanted to have lazy evaluation of `attachTo.element`, you had to use `beforeShowPromise` and modify the options accordingly. With this change, you can use a callback instead.

## Talking points and things to consider:

**1. Should we get rid of `Step.isCentered`?**
`Step.isCentered` is a problematic method. It used to parse `attachTo` options on every call, which isn't good design because we cannot ensure the resolver result will be the same. `attachTo.element` might exist for the initial call, and be gone for another one. What more, it causes trouble in other parts of the library (see 2.), overall making the code uglier. I don't really see the reason for this method to be public, or exist at all - I have created a pure alternative for internal use (`shouldCenterStep`). I left `Step.isCentered` deprecated for now - should we just remove it?
**2. Should `Step._getResolvedAttachToOptions` throw an Error if accessed before `attachTo` evaluation?**
This would be much cleaner, but would prevent `Step._scrollTo` from being called before `_show`, and ideally also require the removal of `Step.isCentered`. Both of these make a lot of sense, `Step.isCentered` is pointless as explained in 1., and `Step._scrollTo` does not need to be called internally before the show phase. This change would make the lazy evaluation logic a lot cleaner. Resolving `attachTo.element` right before a step is rendered makes a lot of sense architecturally, I can't see any other tour lifecycle phase when this would be needed.
**3. This is *potentially* a breaking change.**
Because we memoize the resolver result, this is potentially a breaking change. The reason being that we only invoke `parseAttachTo` once, meaning that for selector strings the corresponding element is only looked for once for every  `Step._show`. Previously, if you supplied a selector string to `attachTo.element`, `parseAttachTo` (and therefore `document.querySelector`) would be called approximately 4 times for every `Step._show`. Of course, this could only become an issue in horribly-written user code, where the programmer toured highly dynamic content, relied on string selectors while doing so, and accepted all the errors being printed in the console, but it's a behaviour change and potential issue nonetheless.
**4. If this qualifies for a major release, maybe we could go for points 1. and 2.?**
It would really make this much more robust, and would only require minor modifications to tests. I would be willing to make those changes.

Super interested to hear your opinion on this. This is an important feature for me, as I'm building a highly-interactive tour of very dynamic content, I would love to see it implemented this way or another. Looking forward to cooperate to see this through.

Related: https://github.com/shipshapecode/shepherd/issues/1177

### DIFF
--- a/docs-src/tutorials/02-usage.md
+++ b/docs-src/tutorials/02-usage.md
@@ -164,7 +164,7 @@ created.
   - `Function` to be executed when the step is built. It must return one the two options above.
 - `title`: The step's title. It becomes an `h3` at the top of the step.
 - `attachTo`: The element the step should be attached to on the page. An object with properties `element` and `on`.
-  - `element`: An element selector string or a DOM element.
+  - `element`: An element selector string, a DOM element, or a function (returning a selector, a DOM element, `null` or `undefined`). 
   - `on`: The optional direction to place the Popper tooltip relative to the element.
     - Possible string values: 'auto', 'auto-start', 'auto-end', 'top', 'top-start', 'top-end', 'bottom', 'bottom-start', 'bottom-end', 'right', 'right-start', 'right-end', 'left', 'left-start', 'left-end'
 
@@ -175,9 +175,13 @@ const new Step(tour, {
 });
 ```
 
-If you don’t specify an attachTo the element will appear in the middle of the screen.
+If you don’t specify an `attachTo` the element will appear in the middle of the screen. The same will happen if your 
+`attachTo.element` callback returns `null`, `undefined`, or a selector that does not exist in the DOM.
+
 If you omit the `on` portion of `attachTo`, the element will still be highlighted, but the tooltip will appear
 in the middle of the screen, without an arrow pointing to the target.
+
+If the element to highlight does not yet exist while instantiating tour steps, you may use lazy evaluation by supplying a function to `attachTo.element`. The function will be called in the `before-show` phase.
 - `beforeShowPromise`: A function that returns a promise. When the promise resolves, the rest of the `show` code for
 the step will execute. For example:
   ```javascript
--- a/landing/css/welcome.css
+++ b/landing/css/welcome.css
@@ -377,7 +377,6 @@ pre {
 button {
   background-color: transparent;
   background-image: none;
-  padding: 0;
 }
 
 /**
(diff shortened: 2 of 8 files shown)


DboyRedd (changes requested):
La policía de las provincias

rwwagner90:
@zefj I forget where we landed on this. I need to review, but have been very busy.

rwwagner90:
@xiwcx @chuckcarpenter I could use some help evaluating this and getting it over the finish line. I definitely like the idea here, but unsure if we want the changes to centered stuff and such.

chuckcarpenter:
@zefj I think if we do a major version bump, no reason not to do as you suggested. 

> I have created a pure alternative for internal use (shouldCenterStep). I left Step.isCentered deprecated for now - should we just remove it?

Your suggestions overall seem cleaner and more performant. Sorry we've been delayed in collaborating, but happy to take next steps to get this in and released. Are you using a fork in your app and have you had any issues with your dynamic tour?

rwwagner90:
@chuckcarpenter happy to defer to you here. I'm mostly in favor of these changes, I just don't have the time to try to grok all of it. I was a little concerned about removing the centered stuff and memoization of the element. I like having the fallback to center if the element isn't there, and I kind of like having the element calculated on the fly each time we call `show`. 

Although, I don't know if my concerns are warranted or not, and I do agree that adding the ability to use a function will be super nice, so people do not have to rely on `beforeShowPromise` as much.

rwwagner90:
@chuckcarpenter @m1073496 can you all work on getting this over the finish line please?

rwwagner90:
Implemented in #1930

PR #1200, opened 2020-10-15
Add div over the clipped region when modalIsVisible is enabled

alexsherekin:
### Reasoning

In `modalIsVisible` mode I would like to be able to apply some custom styling to the highlighted element (e.g. add a ripple effect or change the background color of this region). Currently it's not possible because there is no html element which covers this region.

This PR adds a rectangular element over the clipped region which can be customized with CSS.

### DIFF
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -7,8 +7,17 @@
   let modalIsVisible = false;
   let rafId = undefined;
   let pathDefinition;
+  let pathRect;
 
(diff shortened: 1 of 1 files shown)


rwwagner90:
@alexsherekin I don't think we're interested in adding this. I would recommend instead of forcing Shepherd to add some kind of element where the hole is, you should wrap your element in a div of your desired dimensions and use that as the target.

alexsherekin:
@rwwagner90 completely understood. I managed to solve the problem using a `shepherd-target` CSS class. Thank you!

PR #1237, opened 2020-11-11
Fix overlay not shown after tour complete/cancelled

Rzr7:
Overlay is not appearing again if tour was completed/cancelled.

Bug reproduction steps:

1. Start tour with `useModalOverlay: true` option
2. Complete tour
3. Try to show any step using `tour.show()`

Solution:

If modal container removed - remove modal object too. Otherwise modal container not appearing again.

### DIFF
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -172,7 +172,9 @@ export class Step extends Evented {
    * Hide the step
    */
   hide() {
-    this.tour.modal.hide();
+    if (this.tour.modal) {
+      this.tour.modal.hide();
+    }
 
     this.trigger('before-hide');
 
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -292,6 +292,7 @@ export class Tour extends Evented {
 
         if (modalContainer) {
           modalContainer.remove();
+          this.modal = null;
         }
       }
     }


rwwagner90:
@Rzr7 I don't think we intended to be able to call `show` without calling `start` first. You would bypass a lot of logic that way.

Rzr7:
> @Rzr7 I don't think we intended to be able to call `show` without calling `start` first. You would bypass a lot of logic that way.

I was calling `show` using dropdown "help" list, so I was needed to show only one tour step.

rwwagner90:
@Rzr7 I believe instead of your changes we should do `this._setupModal();` on `show` if the modal doesn't already exist or something. Basically making sure all the logic from `start` gets called if `show` is called before `start` for some reason.

chuckcarpenter:
@rwwagner90 this is an edge case and getting stale. I suggest we close it and happy to take a new PR based on your suggestions if @Rzr7 would like.

rwwagner90:
@chuckcarpenter I did take a look at implementing a fix for this and didn't fully have the time to devote to it. Would love to work with you and/or @xiwcx and see if we can come up with a good fix.

chuckcarpenter:
fixed by #2793

PR #1262, opened 2020-12-04
👷 Move CI/CD to actions

chuckcarpenter:


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,48 @@
...
+      # Install project dependencies
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: yarn
+
+      - name: Build
+        run: yarn build
+
+      - name: Generate .npmrc
+        run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc
```
rwwagner90, line 45: Is generating a `.npmrc` necessary? I would think there is some sort of npm publish action we could use out of the box and just pass a token?

.github/workflows/test.yml
```diff
@@ -0,0 +1,157 @@
+name: Test
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
rwwagner90, line 9:
```suggestion
  pull_request: {}
```

rwwagner90: I think this should be empty, so it runs on all PRs right?

.github/workflows/test.yml
```diff
@@ -0,0 +1,157 @@
...
+
+      # Install project dependencies
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: yarn
+
+      - uses: bahmutov/npm-install@v1
+        env:
+          CYPRESS_INSTALL_BINARY: 0
+      # confirm there is no Cypress installed
```
rwwagner90, line 81: What's the purpose of all this Cypress setup? Shouldn't we use the Cypress defined in our devDeps instead of using `npx` to install the latest?

.github/workflows/test.yml
```diff
@@ -0,0 +1,157 @@
...
+      - run: npx cypress cache path
+      # should return empty list of installed versions
+      - run: npx cypress cache list
```
rwwagner90, line 139: If we're caching Cypress, shouldn't we only need to do all the installation stuff one time?

rwwagner90:
@chuckcarpenter thanks for all your work on this! I know it was a beast 💪

rwwagner90 (approved):

PR #1332, opened 2021-02-19
properly handle centered modal

### LINKED ISSUE #1303: Step without element does not resize it's overlay on window resize
The step which is not attached to any element and is centered does not resize itself if the browser's width or height grows. Other steps update the overlay correctly. The behavior can be seen in the [official  demo](https://shepherdjs.dev/) in 5th step.

1. Start the tour with width 1200px
image

2. Resize the window to 1600px
image

3. The overlay does not update it's size

Is there any way how to fix this issue?

Thanks in advance

xiwcx:
Fixes #1303

### DIFF
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -59,9 +59,9 @@
         y: y - modalOverlayOpeningPadding,
         r: modalOverlayOpeningRadius
       };
+    } else {
+      closeModalOpening();
     }
-
-    closeModalOpening();
   }
 
   /**
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -36,25 +36,32 @@
 
   /**
    * Uses the bounds of the element we want the opening overtop of to set the dimensions of the opening and position it
-   * @param {HTMLElement} targetElement The element the opening will expose
-   * @param {HTMLElement} scrollParent The scrollable parent of the target element
    * @param {Number} modalOverlayOpeningPadding An amount of padding to add around the modal overlay opening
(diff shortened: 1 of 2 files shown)


src/js/components/shepherd-modal.svelte
```diff
@@ -120,15 +127,23 @@
...
    */
   function _styleForStep(step) {
-    const { modalOverlayOpeningPadding, modalOverlayOpeningRadius } = step.options;
+    const {
+      modalOverlayOpeningPadding,
+      modalOverlayOpeningRadius
+    } = step.options;
 
-    if (step.target) {
+    if (step) {
```
rwwagner90, line 135: It looks like this is the main change? We call positionModal now regardless of whether there is a target? I think the `closeModalOpening` stuff maybe needs another look.

src/js/components/shepherd-modal.svelte
```diff
@@ -36,25 +36,32 @@
...
-        height: (height + (modalOverlayOpeningPadding * 2)),
+        width: width + modalOverlayOpeningPadding * 2,
+        height: height + modalOverlayOpeningPadding * 2,
         x: (x || left) - modalOverlayOpeningPadding,
         y: y - modalOverlayOpeningPadding,
         r: modalOverlayOpeningRadius
       };
     }
+
+    closeModalOpening();
```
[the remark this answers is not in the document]
rwwagner90: I think we should only call this when there is no target, right?

rwwagner90 (changes requested):

src/js/components/shepherd-modal.svelte
rwwagner90, line 135: Do we need the if/else here at all anymore? I think `step` has to be defined to get in this method at all.

rwwagner90 (approved):

PR #1449, opened 2021-05-06
🐛 Add preventOverflow option for tether to false

### LINKED ISSUE #1442: "right" not working as an attachTo direction
I'm setting up a fairly simple installation of ShepherdJS. Just trying to create a tour and attach one of the steps to an element on the left edge of the screen. So I'd think that I was set the `on` property of the `attachTo` object to `right`, hoping that the popup would show up on the right side of the element I'm attaching it to. But it shows up over top of the element instead, with the arrow pointing as if it was on the left of the element. If I use DevTools, I can see the x-placement and data-placement attributes set to `left`. Is that not working correctly? Am I using it wrong?

image

chuckcarpenter:
closes #1442

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "watch": "yarn clean && rollup -c --environment DEVELOPMENT --watch"
   },
   "dependencies": {
-    "@popperjs/core": "2.7.2",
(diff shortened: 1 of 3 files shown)


hannakim91:
✅ Looks good!

PR #1479, opened 2021-05-24
Fix issue #1353 : Complete the tour when skipStep is the lastStep

thomasguittonneau:
Call the complete function on the last step 
Fixes https://github.com/shipshapecode/shepherd/issues/1353

### DIFF
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -328,7 +328,7 @@ export class Tour extends Evented {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


rwwagner90:
@thomasguittonneau this looks good! Would you mind adding a test case as well please?

rwwagner90:
@thomasguittonneau we typically make it a rule to only merge PRs with tests, but I will make an exception here.

PR #1632, opened 2021-09-23
I faced a problem while was trying to install shepherdjs

loldalolwerollnroll:
When I was looking through the documentation, I couldn't find how to get proper shepherd styles along with shepherdjs. I found out that, I can download it from jsdeliver in the beginning of my script. So I suggested, that cdn link for styles should be placed near cdn link for shepherdjs in 01-install.md file


rwwagner90 (approved):
Great addition, thanks!

PR #1815, opened 2022-02-18
Add dynamic text and label properties for buttons

radibit:
Hey folks 👋

This PR:
- adds dynamic `text` and `label` properties for the `button` component
- avoids duplicate code for the `button` component (`getConfigOption` function)
- updates the `button` unit tests
- updates `StepOptionsButton` interface
- updates `usage` docs with the new info about the `button` component

The common case for dynamic text and label button properties is when `Shepherd` is used with an i18n (internationalization) solution and those fields need to be updated on change of the language settings. That has been already in place for the [main text component](https://github.com/shipshapecode/shepherd/blob/master/src/js/components/shepherd-text.svelte#L10) but was missing for the buttons.

### DIFF
--- a/test/unit/components/shepherd-button.spec.js
+++ b/test/unit/components/shepherd-button.spec.js
@@ -94,9 +94,10 @@ describe('component/ShepherdButton', () => {
     });
 
     it('label - funtion', () => {
-      const labelFunction = (label) => label;
+      let label = 'Test';
+      const labelFunction = () => label;
       const config = {
-        label: labelFunction('Test')
+        label: labelFunction
       };
 
       const { container, rerender } = render(ShepherdButton, {
@@ -108,7 +109,7 @@ describe('component/ShepherdButton', () => {
       const button = container.querySelector('.shepherd-button');
       expect(button).toHaveAttribute('aria-label', 'Test');
 
-      config.label = labelFunction('Test 2');
+      label = 'Test 2';
 
       rerender({
         props: {
@@ -164,9 +165,10 @@ describe('component/ShepherdButton', () => {
     });
 
     it('text - function', () => {
-      const textFunction = (text) => text;
+      let text = 'Test';
+      const textFunction = () => text;
       const config = {
-        text: textFunction('Test')
+        text: textFunction
       };
 
       const { container, rerender } = render(ShepherdButton, {
@@ -178,7 +180,7 @@ describe('component/ShepherdButton', () => {
       const button = container.querySelector('.shepherd-button');
       expect(button).toHaveTextContent('Test');
 
-      config.text = textFunction('Test 2');
+      text = 'Test 2';
 
       rerender({
         props: {
--- a/test/unit/components/shepherd-button.spec.js
+++ b/test/unit/components/shepherd-button.spec.js
@@ -93,6 +93,21 @@ describe('component/ShepherdButton', () => {
       expect(button).toHaveAttribute('aria-label', '5');
     });
 
+    it('label - funtion', () => {
+      const config = {
+        label: () => 'Test'
+      };
+
+      const { container } = render(ShepherdButton, {
+        props: {
+          config
+        }
+      });
+
+      const button = container.querySelector('.shepherd-button');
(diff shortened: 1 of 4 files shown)


rwwagner90:
Hi @radibit, thanks for the PR! This is definitely a useful feature. Would you mind also updating the docs to mention these things can now be functions and how to use them, etc?

test/unit/components/shepherd-button.spec.js
```diff
@@ -93,6 +93,21 @@ describe('component/ShepherdButton', () => {
       expect(button).toHaveAttribute('aria-label', '5');
     });
 
+    it('label - funtion', () => {
+      const config = {
+        label: () => 'Test'
```
rwwagner90, line 98: @radibit should this test do something to change the value returned by the function after some time and check the value was updated? I assume that is the dynamic updating you are trying to achieve?

radibit: @rwwagner90 By dynamic, I've meant making the `label` and `text` values as a result of a function call next to the option to be primitives of type `string`. That is also what the unit tests I've added are representing.

The use case I had (and successfully tested with the version of this PR) is an i18n solution which has to re-render the content of the app as a callback for changing the language settings, including the tour steps managed by Shepherd. To achieve that, I had to make sure that all text values are functions connected to that i18n solution which will return me the respected translation.

rwwagner90: @radibit I understand your use case, but I am not sure I understand if these are supposed to update or not. I thought, what we were doing here is using the function instead of a string, so that every time the step was shown, the function would be called, and the text would dynamically change. If this is the case, we should update the tests to have a function that changes what it returns, so we can test that when the function returns something different the step shows it. Does that make sense?

radibit: @rwwagner90 Thank you for the clarification. I've got your point and updated the `button` unit tests for the dynamic `text` and `label` properties.

Ref: [9ca563c](https://github.com/shipshapecode/shepherd/pull/1815/commits/9ca563c72e70cf28495ef96546d8bfd45ea2de31)

test/unit/components/shepherd-button.spec.js
```diff
@@ -93,6 +93,33 @@ describe('component/ShepherdButton', () => {
...
+      config.label = labelFunction('Test 2');
+
+      rerender({
+        props: {
+          config
+        }
+      });
+
+      const buttonUpdated = container.querySelector('.shepherd-button');
+      expect(buttonUpdated).toHaveAttribute('aria-label', 'Test 2');
```
rwwagner90, line 120:
```suggestion
      let label = 'Test'
      const labelFunction = () => label;
      const config = {
        label: labelFunction
      };

      const { container, rerender } = render(ShepherdButton, {
        props: {
          config
        }
      });

      const button = container.querySelector('.shepherd-button');
      expect(button).toHaveAttribute('aria-label', 'Test');

      label = 'Test 2';

      rerender({
        props: {
          config
        }
      });

      const buttonUpdated = container.querySelector('.shepherd-button');
      expect(buttonUpdated).toHaveAttribute('aria-label', 'Test 2');
```

@radibit this is what I meant here, changing the value the function returns. The test was calling the functions manually, so you were actually just setting strings here.

radibit: @rwwagner90 You're absolutely right 🙈 My mistake since the tests were indeed obsolete in that form. I've updated both test suites with the last commit [59c323e](https://github.com/shipshapecode/shepherd/pull/1815/commits/59c323ecc7ec68393debfb193c3df2b39ffe5d65)

rwwagner90 (approved):
This looks good! Thanks for making those changes and thanks for the PR.

radibit:
Hey @rwwagner90 sorry to bother you but can I ask if there is a release in planning for the next version of Shepherd library which will include the change completed with this PR?

ℹ️ Unfortunately, there isn't a way to point in my `package.json` to the base (master) branch of Shepherd since the distribution step is part of the npm publishing process.

rwwagner90:
@radibit I released the other day. Sorry for the delay!

PR #1833, opened 2022-03-02
Add modal overlay opening x,y offset settings.

timodwhit:
Add the ability to offset the modal's overlay opening by an x,y value
for elements that are positioned uniquely on the screen. As an example:
This is useful for items that are placed inside an iframe that is in
a fixed position on the screen.


rwwagner90:
@timodwhit looks like we have a failing test here. Mind taking a look?

rwwagner90:
@monshan would you be able to look into this?

monshan:
@rwwagner90 fixed the tests and rebased with most recent shepherd, should be good to go

rwwagner90:
Awesome, thanks @monshan!

rwwagner90:
@monshan would it be possible to add some tests for these new options? I don't think new tests were added here right?

monshan:
@rwwagner90 no tests added by Tim but sure I could write some that cover the feature, my first thought is some cypress that can offset in all 4 directions?

rwwagner90:
@monshan yeah, that sounds perfect! Basically just testing at least a couple different values and making sure it moves.

rwwagner90:
@monshan any updates on those tests? Would love to merge this soon 😄

monshan:
@timodwhit 
Testing the feature will produce the following:
- Left: No offset, modal fills the window
- Right: X-axis offset 25px, modal fills window except for whitespace according to the offset in the bottom left

Can you confirm that this is the desired behavior? Not that the modal offset will remain the same size as the window with adjusted positioning?

Screen Shot 2022-06-22 at 10 46 13 AM

timodwhit:
@monshan I'm sorry. I'm not sure I understand the question or the use case.

EmNicholson93:
@timodwhit @monshan is this issue resolved or are there still lingering questions regarding the tests for this feature?

DerekJarvis:
> @timodwhit @monshan is this issue resolved or are there still lingering questions regarding the tests for this feature?

I believe @monshan asked a question about the purpose of the PR and was awaiting a response. I found this PR because I was looking for this exact feature. I believe the purpose of this feature is to allow the overlay SVG that is generated to offset from the target element. This is useful in cases where the current CSS does not have margins around them and the overlay is very tight. It allows Shepherd to generate a better looking overlay without needing to rework the website's CSS.

timodwhit:
Sorry for the delay in the response and the long gap. 

Yes, @monshan I believe your tests are correct.

chuckcarpenter:
This looks to have been added and handled in #2551

PR #1837, opened 2022-03-05
Separate modal overlay opening element option

### LINKED ISSUE #1836: Separate selectors for modal cutout target and popper target
I'm using the modal overlay in my Shepherd tour, but I'd like to point the tooltip at a specific element that's inside the cutout area. Here's an example, I'd like the tooltip to point directly at the orange button. I tried using `left-start` to move it to the top, but unfortunately no amount of configuration will get the arrow to point directly at the button:

Screen Shot 2022-03-03 at 3 06 59 PM

It'd be great if `attachTo` had an optional second element attribute for either popper target, or the overlay cutout target, so that they can be two separate elements. By default they should both be the same. Not sure if you'd rather an option for each, or go with just one or the other.

wbobeirne:
Closes #1836

* Adds `modalOverlayOpeningElement` option for specifying an `HTMLElement` or string selector for what the modal should target
  * Added documentation for this too
  * Added a cypress test for this new option, but in order to test this, I also added a `getOpeningProperties` method to modal so I could confirm the dimensions matched up with the desired element. Let me know if you'd rather this test be written in a different way that didn't require changing the API of modal.
* Refactored a little in `general` to reuse a `parseElementOrSelector` function for the new option and `attachTo.element`
  * Added a unit test for new `parseElementOrSelector` function


wbobeirne:
Just tried the failing tests locally and they're passing for me. What's the best way to get more info out of CI? I couldn't find a way to get at any specific error logs or the screenshots.

rwwagner90:
@wbobeirne you can see the failing tests by clicking details next to them. 

Seems like the test for `undefined` attachTo is broken https://github.com/shipshapecode/shepherd/runs/5489452652?check_suite_focus=true

rwwagner90:
@wbobeirne would you be able to rebase this with master? We've made some changes, so perhaps tests will work again.

monshan:
@rwwagner90 I looked into this but @wbobeirne 's fork is password protected so I can't push my rebase to his branch, lmk your thoughts if its worth cloning and opening a new PR

rwwagner90:
@monshan looks like we have some failing tests here still. How are things looking?

monshan:
@rwwagner90 I thought the scope of this one would just be to rebase it which went through but still not passing. If I remember right its because this change conflicts with the Shepherd 10 upgrade because now the `attachTo` takes a function.

rwwagner90:
> @rwwagner90 I thought the scope of this one would just be to rebase it which went through but still not passing. If I remember right its because this change conflicts with the Shepherd 10 upgrade because now the `attachTo` takes a function.

Ah yes, that sounds correct.

@wbobeirne could you please refactor this to be compliant with the new `attachTo.element` supporting functions?

rwwagner90:
@EmNicholson93 I think this needs some refactoring to be compliant with the new `attachTo.element` API. Can you please take a look?

PR #1930, opened 2022-05-24
Implement lazy evaluation of attachTo.element

monshan:
# Internal review for #1187 to evaluate lazy attachment of `attachTo.element`

### Changes:
- Remove 
  - ```Step.isCentered()```
- Write unit tests to evaluate ```shouldCenterStep```
- Write cypress tests to evaluate ```parseAttachTo``` binding with multiple elements

---

### Response to Considerations:

**1. Should we get rid of `Step.isCentered` and is `shouldCenterStep` an appropriate alternative?**

- `Step.isCentered` internally approved for removal
  - other methods do not deeply depend on `isCentered`
  - including `shouldCenterStep` to maintain the ability to evaluate the resolved attachTo options will not destroy the feature for any users that are currently or will want to use it in the future
- `shouldCenterStep` maintains and improves upon `Step.isCentered`
  - Testing written in `general.spec.js` to cover reasonable range of potential forms of resolved attachTo options
  - `shouldCenterStep` was modified to handle `null` and `undefined` results

**2. Should `Step._getResolvedAttachToOptions` throw an Error if accessed before attachTo evaluation?**

- Regarding 'Step._scrollTo' being called before `_show`
  - Found in testing that it does not destroy the feature and the handler is still called regardless of attachTo resolution timing
  - However will not produce likely desired results if called before `_show` should dynamic elements be added and removed per step options
  - More of a user behavioral change, recommended to disclose in the documentation
- Should `Step._getResolvedAttachToOptions` throw an error if accessed before the `attachTo` evaluation
  - Would only be material in the case that we would want to internally access the resolved attachTo options from any step other than the current step
  - `setupTooltip` will call this method to also determine a tooltip target, but as mentioned above this is a risk if we are concerned about surrounding steps while on the current step and that we call `setupTooltip` purely

**3. Is lazy attachment potentially a breaking change?**

- The main concern is there is only one opportunity to bind the step to the target vs what could be previously multiple opportunities to bind. This is especially important with dynamic content that may not exist at the beginning of the tour but be generated at the step of interest.
- Further cypress tests were written to check the behavior which include the following:
  - If by chance the attachTo element definition applies to multiple dynamic elements:
    - Only the first element matching the attachTo element definition will be assigned as the target
    - Any other elements that match the element definition if deleted will have no effect on the step as long as the target remains
    - If this targeted element is deleted after the show phase, the step focus and tooltip will still render normally as only the target has been removed
    - AttachTo options (resolved) are the same as non-dynamic content
  - If the whole step is removed (deleted), the attached element is also deleted from the DOM and cannot be found
  - If the whole step is not removed (deleted), the attached element still exists and can be found at any later point in the tour
- Major conclusions:
  - The binding is still successful regardless of the number of matching dynamic elements but only the **first** matching element will be considered the target
  - Even if the target is removed for any reason, as long as it happens post-show of the step the tour can continue as expected
  - Could be a breaking change for users that are binding to a single dynamic element and cannot render it before the show phase of the step in question even using the attachTo option element callback

**4. Would this qualify as a major release?**

- Though the breaking scenarios are niche at bare minimum the following is recommended:
  - Deprecation or removal of `Step.isCentered` (removal done)
  - A footnote in the docs that `._scrollTo` should only be used after `._show` for accurate behavior
  - The lazy `attachTo` can be accepted as long as the above considerations are approved

### DIFF
--- a/docs-src/tutorials/02-usage.md
+++ b/docs-src/tutorials/02-usage.md
@@ -166,7 +166,7 @@ created.
   - `Function` to be executed when the step is built. It must return one the two options above.
 - `title`: The step's title. It becomes an `h3` at the top of the step.
 - `attachTo`: The element the step should be attached to on the page. An object with properties `element` and `on`.
-  - `element`: An element selector string or a DOM element.
+  - `element`: An element selector string, a DOM element, or a function (returning a selector, a DOM element, `null` or `undefined`). 
   - `on`: The optional direction to place the Popper tooltip relative to the element.
     - Possible string values: 'auto', 'auto-start', 'auto-end', 'top', 'top-start', 'top-end', 'bottom', 'bottom-start', 'bottom-end', 'right', 'right-start', 'right-end', 'left', 'left-start', 'left-end'
 
@@ -177,9 +177,13 @@ const new Step(tour, {
 });
 ```
 
-If you don’t specify an attachTo the element will appear in the middle of the screen.
+If you don’t specify an `attachTo` the element will appear in the middle of the screen. The same will happen if your 
+`attachTo.element` callback returns `null`, `undefined`, or a selector that does not exist in the DOM.
+
 If you omit the `on` portion of `attachTo`, the element will still be highlighted, but the tooltip will appear
 in the middle of the screen, without an arrow pointing to the target.
+
+If the element to highlight does not yet exist while instantiating tour steps, you may use lazy evaluation by supplying a function to `attachTo.element`. The function will be called in the `before-show` phase.
 - `beforeShowPromise`: A function that returns a promise. When the promise resolves, the rest of the `show` code for
 the step will execute. For example:
   ```javascript
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -43,6 +43,7 @@ export class Step extends Evented {
    * If you don’t specify an `attachTo` the element will appear in the middle of the screen. The same will happen if your `attachTo.element` callback returns `null`, `undefined`, or a selector that does not exist in the DOM.
    * If you omit the `on` portion of `attachTo`, the element will still be highlighted, but the tooltip will appear
    * in the middle of the screen, without an arrow pointing to the target.
+   * If the element to highlight does not yet exist while instantiating tour steps, you may use lazy evaluation by supplying a function to `attachTo.element`. The function will be called in the `before-show` phase.
    * @param {string|HTMLElement|function} options.attachTo.element An element selector string, DOM element, or a function (returning a selector, a DOM element, `null` or `undefined`).
    * @param {string} options.attachTo.on The optional direction to place the Popper tooltip relative to the element.
    *   - Possible string values: 'auto', 'auto-start', 'auto-end', 'top', 'top-start', 'top-end', 'bottom', 'bottom-start', 'bottom-end', 'right', 'right-start', 'right-end', 'left', 'left-start', 'left-end'
--- a/src/js/step.js
+++ b/src/js/step.js
@@ -40,7 +40,7 @@ export class Step extends Evented {
    * });
    * ```
    *
-   * If you don’t specify an attachTo the element will appear in the middle of the screen.
+   * If you don’t specify an `attachTo` the element will appear in the middle of the screen. The same will happen if your `attachTo.element` callback returns `null`, `undefined`, or a selector that does not exist in the DOM.
    * If you omit the `on` portion of `attachTo`, the element will still be highlighted, but the tooltip will appear
    * in the middle of the screen, without an arrow pointing to the target.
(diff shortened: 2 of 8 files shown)


docs-src/tutorials/02-usage.md
```diff
@@ -166,7 +166,7 @@ created.
   - `Function` to be executed when the step is built. It must return one the two options above.
 - `title`: The step's title. It becomes an `h3` at the top of the step.
 - `attachTo`: The element the step should be attached to on the page. An object with properties `element` and `on`.
-  - `element`: An element selector string or a DOM element.
+  - `element`: An element selector string, a DOM element, or a function (returning a selector, a DOM element, `null` or `undefined`).
```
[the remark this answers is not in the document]
monshan: Great catch, just modified and committed

test/unit/step.spec.js
```diff
@@ -467,6 +469,35 @@ describe('Tour | Step', () => {
...
+
+    it('calls the custom handler after before-show promise resolution', () => {
+      let resHandlerAdded = false;
+      const step = new Step('test', {
+        scrollToHandler: () => (resHandlerAdded = true)
+      });
+
+      parseAttachTo(step)
+
+      step._scrollTo();
```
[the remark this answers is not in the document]
monshan: Before-show promise is more clearly defined in the next commit and passes tests with use of `spy`. Can see `Modify before-show promise on scroll handler in step.spec.js` for more details

src/js/step.js
```diff
@@ -43,7 +43,7 @@ export class Step extends Evented {
    * If you don’t specify an attachTo the element will appear in the middle of the screen.
```
rwwagner90, line 43:
```suggestion
   * If you don’t specify an `attachTo` the element will appear in the middle of the screen. The same will happen if your `attachTo.element` callback returns `null`, `undefined`, or a selector that does not exist in the DOM.
```

src/js/step.js
```diff
@@ -43,7 +43,7 @@ export class Step extends Evented {
...
    * If you omit the `on` portion of `attachTo`, the element will still be highlighted, but the tooltip will appear
    * in the middle of the screen, without an arrow pointing to the target.
```
rwwagner90, line 45:
```suggestion
   * in the middle of the screen, without an arrow pointing to the target.
   * If the element to highlight does not yet exist while instantiating tour steps, you may use lazy evaluation by supplying a function to `attachTo.element`. The function will be called in the `before-show` phase.
```

test/cypress/integration/test.acceptance.js
```diff
@@ -1,5 +1,6 @@
 import setupTour from '../utils/setup-tour';
 import { assert } from 'chai';
+import { doc } from 'prettier';
```
rwwagner90, line 3: Was this added by mistake possibly? I'm not seeing it used anywhere.

test/cypress/integration/test.acceptance.js
```diff
@@ -103,6 +194,175 @@ describe('Shepherd Acceptance Tests', () => {
...
+          const bazTarget = document.createElement('div');
+          bazTarget.setAttribute('class', 'baz');
+          bazTarget.setAttribute('id', 'baz');
+          document.querySelector('[data-test-hero-including]').appendChild(bazTarget);
+
+          const quxTarget = document.createElement('div');
+          quxTarget.setAttribute('class', 'baz');
+          quxTarget.setAttribute('id', 'qux');
+          document.querySelector('[data-test-hero-including]').appendChild(quxTarget);
+          debugger
```
rwwagner90, line 280:
```suggestion
```

test/unit/step.spec.js
```diff
@@ -575,4 +627,124 @@ describe('Tour | Step', () => {
...
+
+      instance.start();
+
+      expect(instance.getCurrentStep().isOpen()).toBe(true);
+      // Subsequent calls to the getter return the same object
+      const result1 = instance.getCurrentStep()._getResolvedAttachToOptions();
+      const result2 = result1;
+      expect(result1).not.toBeNull().then(() => {
+        instance.next();
+        expect(result1).toBe(result2);
```
rwwagner90, line 727:
```suggestion
        const result2 = instance.getById('step1')
        expect(result1).toBe(result2);
```

test/unit/step.spec.js
```diff
@@ -575,4 +627,120 @@ describe('Tour | Step', () => {
...
+          {
+            text: 'step 2',
+            attachTo: { element: () => {}, on: 'auto' }
+          }
+        ]
+      });
+
+      instance.start()
+
+      expect(instance.steps[1]._getResolvedAttachToOptions).toBeTruthy()
```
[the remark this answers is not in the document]
monshan: Replaced with the following:
`expect(instance.getById('step2')._getResolvedAttachToOptions()).toEqual({ element: undefined, on: 'auto'})`

PR #1940, opened 2022-06-01
Update 02-usage.md, note about step header.

do-me:
There's no clue in the docs how to create a step header. The cancel icon only appears if a title was set before, hence the additional note.


rwwagner90:
@do-me that is not correct. The `cancelIcon` is supported without a title. Please see https://github.com/shipshapecode/shepherd/blob/695936b3b51442b2aa2cf9985a528cf4d0b959f7/src/js/components/shepherd-header.svelte#L39

You can also see the first step of our demo has no title, but has a cancelIcon https://shepherdjs.dev/

PR #1942, opened 2022-06-03
Fix chrome opacity bug

### LINKED ISSUE #1941: Background is opaque while popup is shown
Hi
We are able to see opaque backgrounds occasionally while implementing shepherd tour in our project. We are able to get similar scenarios in the demo projects as well. The issue gets automatically fixed after a scroll is done.
Attaching the video of the issue. The issue is happening in chrome. 

Any ideas on how to solve this problem?

https://user-images.githubusercontent.com/23118039/171850337-caa1c748-1f10-4458-a210-f2f598286f55.mp4

https://user-images.githubusercontent.com/23118039/171850340-50f59ec9-22ac-4b42-a6b4-693f98bb5afe.mov

monshan:
### Problem:

Shepherd modal shows opaque (black) background at load in Chrome (not observed in other browsers), but returns to 50% opacity at scroll

### Findings:

- Isolated to the `.shepherd-modal-overlay-container.shepherd-modal-is-visible` css block
- Removing the transition specifications returns to expected 50% opacity at load
- Looked into handling for css transitions in the more recent versions of chrome
  - Reference: [Fixing CSS Transitions in Google Chrome - Ben Gillbanks](https://www.binarymoon.co.uk/2014/02/fixing-css-transitions-in-google-chrome/)

### Solution:

- Add `transform` to css selector to force hardware accelerated transition with value of `0` to have no visual effect other than forcing the hardware transition

Closes #1941

### DIFF
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -231,6 +231,7 @@
     height: 100vh;
     opacity: 0.5;
     transition: all 0.3s ease-out, height 0s 0s, opacity 0.3s 0s;
+    -webkit-transform: translateZ(0);
   }
 
   .shepherd-modal-overlay-container.shepherd-modal-is-visible path {
--- a/src/js/components/shepherd-modal.svelte
+++ b/src/js/components/shepherd-modal.svelte
@@ -231,7 +231,7 @@
     height: 100vh;
     opacity: 0.5;
     transition: all 0.3s ease-out, height 0s 0s, opacity 0.3s 0s;
-    -webkit-transform: translateZ(0);
+    transform: translateZ(0);
   }
 
   .shepherd-modal-overlay-container.shepherd-modal-is-visible path {


src/js/components/shepherd-modal.svelte
```diff
@@ -231,6 +231,7 @@
     height: 100vh;
     opacity: 0.5;
     transition: all 0.3s ease-out, height 0s 0s, opacity 0.3s 0s;
+    -webkit-transform: translateZ(0);
```
rwwagner90, line 234:
```suggestion
    transform: translateZ(0);
```

@monshan I wonder if we can use just normal `transform` here? Not sure we need the `webkit` prefix anymore. Did you test this fix and it fixed the issue?

monshan: @rwwagner90 yes when I did the dev build with this webkit prefix it worked for me, just tried the normal transform though and it also works. I'll re-commit shortly 👍

rwwagner90 (approved):

PR #1997, opened 2022-07-21
Fix for smooth scrolling

hrypkema-amplify:
### Description

Whenever a Tour's default step options is configured with `scrollTo: { behavior: 'smooth', block: 'center' }` the step's target may not always scroll into the center of the viewport when stepping through the tour. 

An example of this behavior can be seen on [Shepherd's welcome page](https://shepherdjs.dev)

What causes this? Shepherd’s popover receives focus while the smooth scroll to the target is in mid-transition.

The sequence of events when a tour step is first shown:
1. Smooth-scrolling to the target begins transitioning
2. After a brief delay, focus is set on the popover and [by default will also trigger a scroll](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) to the element
3. Smooth-scrolling to the target was interrupted by this shift in focus

### Changes
- Configure Shepherd’s focus with `preventScroll: true` so that it does not interrupt scrolling whenever focus shifts to the popover element
- Add supporting tests

### Demo

- Issue: https://shepherdjs.dev
- Fix: https://codepen.io/h/full/QWmgLro

https://user-images.githubusercontent.com/95371739/180328543-ee76fd36-39d4-490c-95fe-f6bef54e0421.mp4

### DIFF
--- a/test/cypress/integration/test.acceptance.cy.js
+++ b/test/cypress/integration/test.acceptance.cy.js
@@ -525,9 +525,10 @@ describe('Shepherd Acceptance Tests', () => {
 
         cy.get('.hero-followup')
           .then((element) => calculateCenteredScrollTop(element[0]))
-          .then((scrollTop) =>
-            cy.window().its('scrollY').should('equal', scrollTop)
-          );
+          .then((scrollTop) => {
+            const plusOrMinusPx = 10;
+            cy.window().its('scrollY').should('be.closeTo', scrollTop, plusOrMinusPx)
+          });
       });
     });
   });
--- a/test/cypress/integration/test.acceptance.cy.js
+++ b/test/cypress/integration/test.acceptance.cy.js
@@ -493,6 +493,42 @@ describe('Shepherd Acceptance Tests', () => {
         tour.next();
         cy.document().get('body').should('have.prop', 'scrollTop').and('eq', 0);
       });
+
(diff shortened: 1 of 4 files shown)


rwwagner90 (approved):
Thank you so much for the incredibly detailed PR with description and tests. This is awesome!

hrypkema-amplify:
@rwwagner90 Thank you for the approval! Taking a look at the test runs and will try get the build green shortly

test/cypress/integration/test.acceptance.cy.js
```diff
@@ -493,6 +493,43 @@ describe('Shepherd Acceptance Tests', () => {
...
+        tour.start();
+        tour.next();
+        tour.next();
+        tour.next();
+
+        cy.get('.hero-followup')
+          .then((element) => calculateCenteredScrollTop(element[0]))
+          .then((scrollTop) => {
+            const plusOrMinusPx = 10;
+            cy.window().its('scrollY').should('be.closeTo', scrollTop, plusOrMinusPx)
```
hrypkema-amplify, line 530: ^ Had to adjust this test to include some margin of difference, since Firefox is capable of scrolling very precisely to `431.616...`

Shepherd Acceptance Tests -- Step options -- scrolling -- scrollToscrollIntoViewOptions scrolls with options (failed)

PR #2037, opened 2022-08-26
Replace popperJS with Floating UI

### LINKED ISSUE #2013: Upgrade from popperjs to floating-ui
Hello,

The popperjs library kinda "rebranded" to https://github.com/floating-ui/floating-ui, are there any plan to make the switch for shepherd? 

In Drupal and we're going to [update our code to use floating-ui instead of popperjs](https://www.drupal.org/project/drupal/issues/3301545). And since we use shepherd and shepherd uses popperjs it'd be great to not have two different libraries for that.  

Are there any plans or interest in moving over to floating-ui? in a few weeks we should be able to help with making a PR to switch, before spending time on it I just wanted to make sure it's a welcome change :)

theodoreb:
Fixes #2013

This PR removes popperjs dependency and replace it with floating UI.

### Configuration similarities
For `popperOptions` and `floatingUIOptions`, the `placement` and `strategy` options are the same.

### Configuration differences
There is no equivalent to popperjs `onFirstUpdate` option. Floating UI is low-level and the lifecycle is up to the implementation. In this PR the equivalent would be to add a `.then()` to the promise in `floating-ui.js:setPosition` function. See the implementation for `focusAfterRender`.

#### floatingui middleware
Middleware are the equivalent of popperjs modifiers with a few differences. Middleware are used to modify coordinates for the tooltip. They stack on each others, one middleware using the return from the previous middleware. Order is left to the user (hence the removal of the `phase` option) and they are executed in the order they come in the middleware array.

#### Step class API changes

The `Step.show` method always returns a promise:

* when `beforeShowPromise` is set and returns a promise, 
* when `beforeShowPromise` is set and does not return a promise, 
* and when `beforeShowPromise` is not set.

#### Binding between step.js and the tooltip/positioning library.

For floating UI (or any other library if a change is required) there are 3 functions to implement that are used in `step.js`: 

* `setupTooltip`: Called when creating/showing a step, is responsible for positioning/showing/binding events to keep the tooltip visible.
* `destroyTooltip`: Called when destroying a step, is responsible for clearing events/objects used by the tooltip library
* `mergeTooltipConfig`: Called when building a step options, is responsible for merging library-dependent options between the tour's `defaultStepOptions` and the individual step-specific options.
* `getFloatingUIOptions`: Technically exported but only used in tests.


package.json
```diff
@@ -89,6 +90,7 @@
     "release-it": "^15.3.0",
     "release-it-lerna-changelog": "^5.0.0",
     "replace": "^1.2.1",
+    "resize-observer-polyfill": "^1.5.1",
```
rwwagner90, line 93: @theodoreb why was this added? If it is required, does it need to be in `dependencies` instead of `devDependencies`?

theodoreb: it's only for running the tests because floatingui `autoUpdate` uses ResizeObserver and that fails if it's not available.

theodoreb: see https://floating-ui.com/docs/autoUpdate

rwwagner90:
@theodoreb there are tests that use `popperOptions.modifiers` those will need to be updated and we'll need to update docs around modifiers to work with floating UI.

theodoreb:
Thanks for the feedback. In general, how do you want to handle the API change? I don't see a way to automatically transform custom popperjs modifiers into floatingui middleware (they're similar but different). 

What type of situation did you have in mind: 

1. **Clean break**. Release a new major version with floatingui support only, explain how to do the upgrade in the release notes. Remove all support for popperjs-based code. 
2. **Deferred break** Release a new minor version that support both floatingUI and popperjs so that people can update their code on their time, possibly ship with 2 different set of built files depending on the supported positioning library. Say that the popperjs version is deprecated and remove it during the next major release.
3. **BC Compatibility** Release a minor version and somehow provide BC support for all the major popperjs functionality.

Solution 1 is the easiest technically, nothing to duplicate, fixing tests is straightforward but a lot of effort will/should go into the release notes to make it very clear how to update custom popperjs code. This comes at the cost of breaking all shepherd implementations.

I think solution 2 would be the least disruptive for users with a manageable amount of work for maintainers, we'd have to update the tests to handle the 2 different positioning libraries. The work can be less exhaustive for the upgrade since it won't break user code and we can "just" point out to [floatingui migration guide](https://floating-ui.com/docs/migration). 

I have no idea how feasible solution 3 is and I'd rather avoid it if possible as maintaining a BC layer gets messy pretty fast (and adds a whole bunch of tests). but it might be possible to go this way.


Which scenario would you prefer? or do you have something else in mind?

theodoreb:
As far as time goes, I have enough time dedicated to this to implement/help out with either 1 or 2. My personal preference is for 2 as it won't break existing implementations, and I won't have to struggle to write clear documentation :) 

Another option could be to render the positioning part somewhat DIY and define a simple enough interface positioning libraries should follow. Not sure it's worth it but at least I kinda know what it'd take unlink option 3.

From the comment you made earlier seems like you had option 1 in mind, I'll go along with that for now and we can always adjust later when tests are green.

theodoreb:
Ping @EmNicholson93 if you have an idea about which option to go with :)

rwwagner90:
@theodoreb I think we would prefer a clean break and just updating all the way. We don't need backwards compatibility.

theodoreb:
Sounds good, will update the PR tomorrow 👍

test/cypress/integration/destroying-elements.cy.js
```diff
@@ -43,44 +43,41 @@ describe('destroying-elements', () => {
...
-        tour.next();
-
-        // Remove the first element
-        document.querySelector('.first').remove();
+    let initialPosition;
+    let missingStepElementPosition;
+    let finalPosition;
 
-        tour.back();
+    cy.get('[data-shepherd-step-id="first"]').then((stepElement) => {
```
theodoreb, line 50: removed some nesting from the promise chain for readability

test/unit/step.spec.js
```diff
@@ -91,21 +98,22 @@ describe('Tour | Step', () => {
...
-      popperOptions: {
-        modifiers: [{ name: 'offset', options: { offset: [0, 0] } }]
+      floatingUIOptions: {
+        middleware: [stepWithoutNameWithoutIdOffsetMiddleware]
       }
     });
 
-    const beforeShowPromise = new Promise((resolve) => {
-      return setTimeout(() => resolve('beforeShowPromise worked!'), 1000);
+    const beforeShowPromise = () => new Promise((resolve) => {
```
theodoreb, line 106: Changed the nesting, we don't access this promise directly, we're looking at the output of the show() method to get the promise.

test/unit/tour.spec.js
```diff
@@ -446,16 +451,17 @@ describe('Tour | Top-Level Class', function() {
...
+        },
       });
 
       instance.start();
 
-      const popperOptions = setupTooltip(step);
+      const floatingUIOptions = setupTooltip(step);
 
-      expect(popperOptions.modifiers.length).toBe(5);
+      // @todo should be 3 here.
```
theodoreb, line 463: need to dig in what's going on here. should be 3 middleware, the default 2 + the `foo` middleware

landing/css/welcome.css
```diff
@@ -1,5 +1,5 @@
 /*
-! tailwindcss v3.1.7 | MIT License | https://tailwindcss.com
+! tailwindcss v3.1.8 | MIT License | https://tailwindcss.com
```
theodoreb, line 2: side effect of running the build command

src/js/step.js
```diff
@@ -159,13 +155,10 @@ export class Step extends Evented {
    * Triggers `destroy` event
    */
   destroy() {
-    if (this.tooltip) {
-      this.tooltip.destroy();
-      this.tooltip = null;
-    }
+    destroyTooltip(this);
```
theodoreb, line 158: moved this bit in it's own function to match what's happening with `setupTooltip`.

src/js/step.js
```diff
@@ -159,13 +155,10 @@ export class Step extends Evented {
...
 
-    if (isHTMLElement(this.el) && this.el.parentNode) {
-      this.el.parentNode.removeChild(this.el);
+    if (isHTMLElement(this.el)) {
+      this.el.remove();
```
theodoreb, line 161: should be safe with browser support (since floating ui uses async/await that should be available as well)

src/js/utils/floating-ui.js
```diff
@@ -0,0 +1,180 @@
...
+  if (step.cleanup) {
+    step.cleanup();
+  }
+
+  const attachToOptions = step._getResolvedAttachToOptions();
+
+  let target = attachToOptions.element;
+  const floatingUIOptions = getFloatingUIOptions(attachToOptions, step);
+
+  if (shouldCenterStep(attachToOptions)) {
```
theodoreb, line 32: default behavior of floatingui is to center the tooltip, no need to do anything specific here.

theodoreb (commented):
Some more context for a few of the changes

theodoreb:
I think all the code logic pieces are there. 

Styles/attributes names/documentation are not updated yet. Waiting for some feedback on the code/progress before working on those.

src/js/step.js
```diff
@@ -236,12 +233,11 @@ export class Step extends Evented {
    */
   show() {
     if (isFunction(this.options.beforeShowPromise)) {
-      const beforeShowPromise = this.options.beforeShowPromise();
-      if (!isUndefined(beforeShowPromise)) {
-        return beforeShowPromise.then(() => this._show());
-      }
+      return Promise.resolve(this.options.beforeShowPromise()).then(() =>
```
theodoreb, line 236: Wrapping in Promise.resolve makes sure that whatever beforeshowpromise method return we can use it as a promise. If a promise is returned it's unwrapped by Promise.resolve.

src/js/step.js
```diff
@@ -236,12 +233,11 @@ export class Step extends Evented {
...
+        this._show()
+      );
     }
-    this._show();
+    return Promise.resolve(this._show());
```
theodoreb, line 240: This is for consistency so that show() always returns a promise.

src/js/step.js
```diff
@@ -357,7 +353,8 @@ export class Step extends Evented {
         arrow: true
       },
       tourOptions,
-      options
+      options,
+      mergeTooltipConfig(tourOptions, options)
```
theodoreb, line 357: same as setupTooltip/destroyTooltip, moved that code to it's own file so we can change all floatingUI related options in one file.

src/js/utils/floating-ui.js
```diff
@@ -0,0 +1,203 @@
...
+      bottom: 'top',
+      left: 'right'
+    }[placement.split('-')[0]];
+
+    Object.assign(arrowEl.style, {
+      left: arrowX != null ? `${arrowX}px` : '',
+      top: arrowY != null ? `${arrowY}px` : '',
+      right: '',
+      bottom: '',
+      [staticSide]: '-35px'
```
theodoreb, line 157: probably should be a config/CSS rule somewhere.

rwwagner90:
@monshan @EmNicholson93 could you please take a first pass at reviewing this?

rwwagner90:
@monshan @EmNicholson93 could you please take a first pass at reviewing this?

monshan (changes requested):
First pass for review

test/unit/step.spec.js
```diff
@@ -155,28 +163,20 @@ describe('Tour | Step', () => {
...
+          middleware: [defaultOffsetMiddleware, fooMiddleware],
         },
         showOn,
         when
       });
     });
 
-    it('allows the step to override a previously defined modifier', () => {
+    // @todo decide what to do here.
+    it.skip('allows the step to override a previously defined modifier', () => {
```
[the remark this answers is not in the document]
theodoreb: Yeah wanted to talk about this one, since the format of the middleware changed, and that middleware are made to be composable, does it still make sense to allow overrides and instead just add more middlewares to do what's necessary?

theodoreb:
Sorry didn't see the review/notification until today for some reasons!

test/cypress/integration/destroying-elements.cy.js
```diff
@@ -43,44 +43,41 @@ describe('destroying-elements', () => {
...
-          tour.next();
+      tour.next();
 
-          // Create the first element again
-          const first = document.createElement('div');
-          first.className = 'first';
-          first.textContent = 'First';
-          document.body.appendChild(first);
+      // Create the first element again
+      Cypress.$('body').append('<div class="first">First</div>');
```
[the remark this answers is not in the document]
theodoreb: Removed both uses of Cypress.$

theodoreb:
I've updated the merge request description to outline the big code changes. 

I'd like to discuss whether we want to keep the "alterability" of middlewares considering we can stack them as much as we want and they're meant to be stacked that way: https://floating-ui.com/docs/middleware#ordering

test/unit/step.spec.js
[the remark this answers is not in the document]
theodoreb: Unskiped the test this is the only test failure left where I think we might need to discuss options.

rwwagner90:
@theodoreb I think stacking middleware seems reasonable. We just want to make sure people can change the behaviors if they need to.

rwwagner90:
@theodoreb it looks like we have a test failure and a conflict. If you can fix those things up, I can do another pass at reviewing. Thank you for all your work here!

theodoreb:
Still on holidays for a couple of weeks will get to it right afterwards. Thanks!

theodoreb:
updated the test to check that middleware stacking works as expected

rwwagner90 (approved):
This looks good to me, thank you so much @theodoreb for the hard work!

rwwagner90:
@theodoreb it looks like we still need to document these changes. The docs do not mention how to interact with floating UI and we still have references to `popperOptions` etc. Would you be able to help update things please?

theodoreb:
@rwwagner90 happy to, opened #2118 will get to it during the week

PR #2046, opened 2022-08-29
Fix calls to setTimeout in tests

theodoreb:
When trying to run the test suite I've run into failures on the main branch. That prevent refactoring work so here is what fixed it for me locally.


theodoreb:
I didn't clean the commits because I'd expect them to be squashed with the commit message you're ok with.

rwwagner90 (approved):

PR #2048, opened 2022-08-31
adjust staleLabel to better articulate its purpose

EmNicholson93:
I'd like to propose we change the label staleBot applies on issues that get old with no movement. `wontFix` implies that that the issue is not something worth doing anything about. Changing the applied tag to `stale` will better express that the issue has simply fallen to the wayside for owners, contributors and issue OP

I've added a `stale` label so as long as there are no objections this is a one line change

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -7,7 +7,7 @@ exemptLabels:
   - pinned
   - security
 # Label to use when marking an issue as stale
-staleLabel: wontfix
+staleLabel: stale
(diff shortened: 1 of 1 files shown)


rwwagner90:
@EmNicholson93 `wontfix` is by design. That way when stalebot eventually closes the issue, it is tagged correctly to not be picked up in the changelog. That being said, we could remove the bot entirely. I'm not sure we really need it anymore.

PR #2049, opened 2022-09-01
Noop ssr

EmNicholson93:
I believe this resolves issue #961. I wasn't able to figure out a way to test the NOOP, I thought about trying to spoof window as undefined but struggled to find a way to do that. Any thoughts on getting tests in place for this would be welcomed.

### DIFF
--- a/src/js/tour.js
+++ b/src/js/tour.js
@@ -13,6 +13,8 @@ import ShepherdModal from './components/shepherd-modal.svelte';
 
 const Shepherd = new Evented();
 
+const noop = () => {};
+
 /**
  * Class representing the site tour
  * @extends {Evented}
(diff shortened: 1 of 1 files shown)


theodoreb:
Should it be throwing an error instead? 

Seems like none of the object methods will be available and if there is code using shepherd API it'll fail in possibly obscure way. if it fails when trying to use or even load the thing on the server it might be clearer what's supposed to happen maybe?

chuckcarpenter:
@theodoreb I don't believe it should through an error on the server, but am open to alternatives in that scenario. Wouldn't you just load Shepherd programmatically once on the client? Unclear what you'd need in SSR from the API?

theodoreb:
All right read the associated issue again. Don't mind me i misunderstood 🙂

theodoreb:
Although changing the constructor feels like it's a little late? Maybe it could be replacing Tour (and Step?) In index.js, that would seem more appropriate to me

rwwagner90:
@EmNicholson93 perhaps this would be helpful for mocking the window in Jest? https://thewebdev.info/2022/02/24/how-to-mock-the-javascript-window-object-using-jest/

As for how to mock in cypress, perhaps we could use sinon? Not sure if cypress has something built in or not.

rwwagner90:
@EmNicholson93 where are we with mocking some of this so we can test it?

EmNicholson93:
@rwwagner90 Planning to get to this either tomorrow or Monday

chuckcarpenter:
```
it('does not start a tour when window is undefined', () => {
      const windowSpy = jest.spyOn(window, 'window', 'get');
      
      windowSpy.mockImplementation(() => undefined);
      instance = new Shepherd.Tour({ defaultStepOptions });

      expect(instance).toBeUndefined();
    });
```

That can be done in the unit tests and cleaner in it's own file I think. I also believe I agree with @theodoreb that it should be sooner in shepherd.js with something like:

```
const isServerSide = typeof window === 'undefined';
const noop = () => undefined;

if (isServerSide) {
  Object.assign(Shepherd, { Tour: noop });
} else {
  Object.assign(Shepherd, { Tour, Step });
}
```

PR #2068, opened 2022-09-15
Do nothing when running server side

### LINKED ISSUE #961: noop when SSR
We should look at making this library a noop when not running on the client to avoid SSR errors for people.

chuckcarpenter:
Closes #961


rwwagner90:
@chuckcarpenter @EmNicholson93 I got the jest tests passing, but seems like I broke the cypress ones. We needed to make our noop a class, so calling `new Shepherd.Tour` would work. I'm unsure what the best place to do these checks is. We may want to move it back to checking in the constructor, but open to ideas here.

EmNicholson93:
@rwwagner90 @chuckcarpenter I believe this is good to go now. Removing the NoOp export allows the cypress tests to pass and expecting Tour to simply return truthy gets the jest test passing, no need to try and bring in NoOp

PR #2083, opened 2022-09-30
feature: add support for multiple elements highlight

IArny:
I would propose feature to support multiple elements highlight. All elements selected with attachTo.element will be highlighted if attachTo.multiple is true. I added multiple flag to attachTo field to keep old behavior. I'd appreciate your feedback for this PR.

Also this PR can resolve https://github.com/shipshapecode/shepherd/issues/525 issue


Krumil:
This will be a life-saver for me! Thanks

rwwagner90:
@IArny would you be able to rebase and fix conflicts please? Thanks!

rwwagner90:
@IArny I'm so sorry, but it looks like we have conflicts again. Could you rebase one more time please?

IArny:
@rwwagner90 I've fixed conflicts and tests, could you check it agait, please

IArny:
@rwwagner90 Could you review my PR, please

IArny:
Hey, is anyone there? Would appreciate this to be merged to master @rwwagner90

RobbieTheWagner:
@IArny I finally got a chance to take a look at this. I think it might make more sense to have `attachTo` be an array of objects with `{element, on}` like it currently is, just would accept an array instead of just a single element. If we wanted to keep the old API as well, we could use something like `isArray` to check if the option passed is an array, etc. What do you think?

IArny:
@RobbieTheWagner thanks for review, I think that `attachTo` as an array would be useful feature, but we can leave current suggestion as an option too. Maybe you can merge current PR and I will make `attachTo` as an array in next PR? A bit difficult to support this PR as it contains some linting updates.

RobbieTheWagner:
@IArny I don't think we should merge this PR, unfortunately. I think basically we should do:

```js
if(Array.isArray(attachTo) {
  // loop through the attachTo options and do all the attachTo things
} else {
// Just do what we already do when it's a single element
}
```

I do not like that your implementation requires a root element selector and selects all elements under it when you pass `multiple`. I would rather be explicit about all the elements you want highlighted.

juanitoddd:
This will be much a nice feature !

500Foods:
So it looks like this is almost ready for release? Is that the case?

davidperezvw:
Are we gonna see this change on release any soon?

500Foods:
Lots of repo activity! Is this in the works now?

RobbieTheWagner:
> @IArny I don't think we should merge this PR, unfortunately. I think basically we should do:
> 
> ```js
> if(Array.isArray(attachTo) {
>   // loop through the attachTo options and do all the attachTo things
> } else {
> // Just do what we already do when it's a single element
> }
> ```
> 
> I do not like that your implementation requires a root element selector and selects all elements under it when you pass `multiple`. I would rather be explicit about all the elements you want highlighted.

This was the last place we left off. This is not planned currently, but if someone wanted to spike it out, we would definitely accept PRs!

RobbieTheWagner:
Highlighting multiple things was implemented in https://github.com/shipshapecode/shepherd/pull/2995

PR #2168, opened 2022-12-05
add funtion support for confirmCancel

taozhiyu:
demo

Promise function:
```js
// TypeScript
{
  ...
  confirmCancel: () => {
    return new Promise<boolean>((resolve) => {
      // my dialog component
      dialog
        .open(
          "Are you sure you want to stop the tour?",
          "You can watch it again in setting paddle"
        )
        .then((res: boolean) => {
          resolve(res);
        });
    });
  },
  ...
}
```
https://github.com/shipshapecode/shepherd/blob/1076b0c37465cb48968b5f7159e08033cceaab20/test/unit/tour.spec.js#L295-L335
Unit tests have been passed.


taozhiyu:
Can anyone help to see if there is really a problem with this pr?

It doesn't seem to pass the 'cognitive complexity' test, but I don't really know how to reduce it any more.

(In fact, I only added one judgement).

src/js/tour.js
```diff
@@ -129,17 +130,26 @@ export class Tour extends Evented {
...
-        this.options.confirmCancelMessage ||
-        'Are you sure you want to stop the tour?';
-      const stopTour = window.confirm(cancelMessage);
+  async cancel() {
+    if (!this.options.confirmCancel) {
+      this._done('cancel');
+      return;
+    }
+    const type = Object.prototype.toString.call(this.options.confirmCancel);
+    if (type === '[object Function]' || type === '[object AsyncFunction]') {
```
rwwagner90, line 142:
```suggestion
    if (typeof this.options?.confirmCancel === 'function') {
```

Would this work? Seems a bit simpler.

taozhiyu: Thanks, this does make it easier, which I will adopt later, but most importantly, based on the error message reported by the checker as shown below, the root of the error may not be here?

image

Of course, it could also be that find this problem here and then it just quits

src/js/tour.js
```diff
@@ -129,13 +130,19 @@ export class Tour extends Evented {
   /**
    * Calls _done() triggering the 'cancel' event
    * If `confirmCancel` is true, will show a window.confirm before cancelling
+   * If `confirmCancel` is a function, will call it and wait for the return value,
+   * and only cancell when the value returned is true
```
rwwagner90, line 134:
```suggestion
   * and only cancel when the value returned is true
```

src/types/tour.d.ts
```diff
@@ -34,6 +34,8 @@ declare class Tour extends Evented {
   /**
    * Calls done() triggering the 'cancel' event
    * If `confirmCancel` is true, will show a window.confirm before cancelling
+   * If `confirmCancel` is a function, will call it and wait for the return value,
+   * and only cancell when the value returned is true
```
rwwagner90, line 38:
```suggestion
   * and only cancel when the value returned is true
```

taozhiyu:
Hi, is there anything else I need to do?

I see that this PR has not been merged, I just don't know if there is anything else that needs to be done.

RobbieTheWagner (approved):

dvopalecky:
Hello, thank you for this feature. @RobbieTheWagner, when do you plan to release a new version in npm with this update? Thank you

RobbieTheWagner:
A new version has been released

PR #2204, opened 2023-01-10
Bugfix: prevent to show spurious warnings in dev panel console

SamyCookie:
warnings occur in Firefox Web dev panel console when some attribute in `Element.style` component are set to `undefined`

image


SamyCookie:
updated for linting

rwwagner90 (approved):
LGTM! Thanks for the PR

PR #2283, opened 2023-03-15
Update welcome.js

aasim-syed:
Here , I did the following changes.

- [x] Added  a comment to explain the purpose of the ready() function.

- [x] Used the ```addEventListener``` method to check for the load event instead of the ```attachEvent``` method, as the latter is no longer recommended.

- [x] Used the ```DOMContentLoaded``` event instead of the load event to ensure that the DOM is ready, but images and other resources may still be loading.

### DIFF
--- a/landing/js/welcome.js
+++ b/landing/js/welcome.js
@@ -180,13 +180,14 @@
     return shepherd;
   }
 
-  function ready() {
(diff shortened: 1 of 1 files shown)


RobbieTheWagner:
We have refactored this and removed the extra checks. Thank you for the PR though!

PR #2399, opened 2023-07-07
Prevent ESC, KEY_RIGHT, KEY_LEFT propagation when keyboardNav is enabled

karendolan:
This pull prevents key down event propagation on ESC, KEY_RIGHT, KEY_LEFT keys in Shepherd, when the Shepherd keyboardNavigation is enabled in a shepherd tour.

The reason: our underlying web application, like may others, has a listener for ESC, KEY_RIGHT, KEY_LEFT key press to enable shortcut actions. When the Shepherd "onboarding" tour is played and the user uses keyboardNav, our underlying web application also receives the KEY_RIGHT, KEY_LEFT, and ESC events and also handles them. This causes havoc and confusion with the underlying web application doing things that the user did not intend while they were only trying to navigate the Shepherd tour.

### DIFF
--- a/src/js/components/shepherd-element.svelte
+++ b/src/js/components/shepherd-element.svelte
@@ -92,16 +92,19 @@
         break;
       case KEY_ESC:
         if (tour.options.exitOnEsc) {
+          e.stopPropagation();
           step.cancel();
         }
         break;
       case LEFT_ARROW:
         if (tour.options.keyboardNavigation) {
+          e.stopPropagation();
           tour.back();
         }
         break;
       case RIGHT_ARROW:
         if (tour.options.keyboardNavigation) {
+          e.stopPropagation();
           tour.next();
         }
         break;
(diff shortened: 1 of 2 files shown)


RobbieTheWagner:
@karendolan thanks for the PR, this looks great! I am a bit confused on the testing coverage though. Can you please explain how it works?

karendolan:
> 



> @karendolan thanks for the PR, this looks great! I am a bit confused on the testing coverage though. Can you please explain how it works?

Hi @RobbieTheWagner I tweaked the test to improve it and added comments. The test added an event listener to the body, i.e. parent container, to test if the keydown event trickles up to it.
- The first test, "keyboard navigation true", is when the keydown event should not get to the parent. Because the user's intent is to navigate Shepherd.
-  In the second test, "keyboard navigation false", the keydown event *should* get to the parent because they keys are not being used to navigate Shepherd, they are being used for their default behavior, which is what ever the parent is doing with them.

PR #2499, opened 2023-10-12
Add section for inserting progress html

sirmspencer:
Add a section in the footer for add a progress bar html.

I don't know svelt so I'm sure this isn't 100% correct, but its the general idea.

#2496

### DIFF
--- a/src/js/components/shepherd-footer.svelte
+++ b/src/js/components/shepherd-footer.svelte
@@ -4,6 +4,15 @@
   export let step;
 
   $: buttons = step.options.buttons;
+  $: progress = step.options.progress;
+
+  afterUpdate(() => {
+    if (isFunction(progress)) {
+      progress = progress();
+    }
+
+    if (progress) element.innerHTML = progress;
+  });
 </script>
 
 <style global>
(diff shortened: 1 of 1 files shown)


chuckcarpenter:
@sirmspencer sorry for the late response, but there are a couple of options for adding optional progress indicators in the [docs](https://shepherdjs.dev/docs/tutorial-04-cookbook.html). Would that meet your use case?

sirmspencer:
> @sirmspencer sorry for the late response, but there are a couple of options for adding optional progress indicators in the [docs](https://shepherdjs.dev/docs/tutorial-04-cookbook.html). Would that meet your use case?

It doesn't because the buttons are already in the footer and there are not enough divs to organize.  I will take one more chance to look to see if I can leave out the shepherd buttons to have then entire footer and add my own buttons.

chuckcarpenter:
Going to close this one because using the cookbook recipes and playing around with further changes I was able to make progress appear in various areas. If you have a codesandbox or similar I'm happy to help figure out the use case otherwise. e.g., 
image

PR #2551, opened 2023-12-13
fix: handle target elements within an iframe

rafiazman:
Corrects the positioning of the overlay cutout when the target element is within an iframe.

Issue:: #1692


chuckcarpenter:
@rafiazman looks like we have one failed unit test here. Awesome work though and I'll try to test some locally as well myself!

rafiazman:
> @rafiazman looks like we have one failed unit test here. Awesome work though and I'll try to test some locally as well myself!

Fixed 👍

PR #2664, opened 2024-03-27
Add extra properties bucket to send to Shepherd Pro

RobbieTheWagner:


chuckcarpenter (approved):

chuckcarpenter:
I do think there should be some defaults about the user that could be added here, e.g., OS, Browser, etc..

PR #2766, opened 2024-05-01
yarn to pnpm spike

RobbieTheWagner:


chuckcarpenter:
renaming isn't going to be enough here, the basic commands don't work. e.g., 
image

RobbieTheWagner:
> renaming isn't going to be enough here, the basic commands don't work. e.g., image

Yeah, I know renaming isn't the only thing we need. Stuff with docker needs to be done which I didn't do here as I don't know what to do there.

chuckcarpenter:
@RobbieTheWagner way before docker is addressed the basic redwood commands don't work, e.g., `pnpm rw dev`

chuckcarpenter:
No longer needed with #2846

PR #2805, opened 2024-05-13
🐛 Fixes import issues reported in 2785

### LINKED ISSUE #2785: TypeScript errors in 12.0.0
After upgrading to Shepherd.js 12.0.0, TypeScript no longer compiles my code that uses `shepherd.js`. It fails with the following error:

> error TS2307: Cannot find module 'shepherd.js' or its corresponding type declarations.

For reference, [Are the Types Wrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io) is reporting additional problems for 12.0.0, compared to 11.2.0:

* https://arethetypeswrong.github.io/?p=shepherd.js%4012.0.0
* https://arethetypeswrong.github.io/?p=shepherd.js%4011.2.0

chuckcarpenter:
closes #2785 

image

The issues with imports were due to `.svelte` imports in the declarations files and that not resolving correctly (workaround for now but just using `Record`) and the need for two declarations files for both cjs and esm.

### DIFF
--- a/shepherd.js/package.json
+++ b/shepherd.js/package.json
@@ -55,8 +55,7 @@
   "dependencies": {
     "@floating-ui/dom": "^1.5.3",
     "deepmerge-ts": "^5.1.0",
-    "idb": "^8.0.0",
-    "rollup-plugin-dts": "^6.1.0"
+    "idb": "^8.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.24.5",
@@ -79,6 +78,7 @@
     "rimraf": "^5.0.5",
     "rollup": "^4.17.2",
     "rollup-plugin-analyzer": "^4.0.0",
+    "rollup-plugin-copy": "^3.5.0",
     "rollup-plugin-filesize": "^10.0.0",
     "rollup-plugin-license": "^3.3.1",
     "rollup-plugin-livereload": "^2.0.5",
--- a/shepherd.js/package.json
+++ b/shepherd.js/package.json
@@ -18,20 +18,19 @@
     "Robbie Wagner <rwwagner90@gmail.com>",
     "Chuck Carpenter <chuck@shipshape.io>"
   ],
+  "main": "dist/shepherd.js",
+  "module": "dist/shepherd.mjs",
   "exports": {
     ".": {
-      "types": "./dist/shepherd.d.ts",
-      "require": "./dist/shepherd.js",
       "import": "./dist/shepherd.mjs",
-      "default": "./dist/shepherd.js"
+      "require": "./dist/shepherd.js"
     },
     "./*": {
-      "types": "./dist/*.d.ts",
-      "require": "./dist/*.js",
       "import": "./dist/*.mjs",
-      "default": "./dist/*.js"
+      "require": "./dist/*.js"
     }
   },
+  "types": "shepherd.d.ts",
   "typesVersions": {
     "*": {
       "*": [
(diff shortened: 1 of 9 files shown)


shepherd.js/package.json
```diff
@@ -18,20 +18,19 @@
...
     },
     "./*": {
-      "types": "./dist/*.d.ts",
-      "require": "./dist/*.js",
       "import": "./dist/*.mjs",
-      "default": "./dist/*.js"
+      "require": "./dist/*.js"
     }
   },
+  "types": "shepherd.d.ts",
```
RobbieTheWagner, line 33: I don't think we want this. We had all the types files in the dist before and we want them to correspond to the files people are importing.

shepherd.js/rollup.config.mjs
```diff
@@ -45,7 +46,18 @@ const plugins = [
     banner
   }),
   filesize(),
-  visualizer()
+  visualizer(),
+  copy({
```
RobbieTheWagner, line 50: I don't think we want to manually copy files. Was there something we needed an `mts` for? I think `mjs` was all correctly tied to the declaration files and working already for me.

shepherd.js/scripts/postprocess.js
```diff
@@ -0,0 +1,7 @@
+const fs = require('fs');
```
RobbieTheWagner, line 1: Manually changing this stuff feels like a hack. I would like to know more about the problems people were experiencing and if it was just with node 16 or lower or what exactly was going on, since all this stuff worked for me in both CJS and ESM projects.

shepherd.js/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
   "compilerOptions": {
-    "target": "es2021",
-    "module": "esnext",
-    "moduleResolution": "bundler",
+    "target": "es2020",
+    "module": "es2020",
+    "moduleResolution": "node",
```
RobbieTheWagner, line 6: I'm not sure we want to change these settings. Shepherd is not supported to be used in node, but in browser only, so it should be `bundler` still I think.

chuckcarpenter: > While few libraries care about supporting a long-past EOL version of Node, maintainers of very popular libraries should be aware that many TypeScript users are still using --moduleResolution node. If they aren’t literally using Node 10, they really should move away from it, but sometimes migrating can be difficult because it introduces new errors caused by incorrect dependency typings—the problem this tool was made to diagnose. Popular libraries often choose to adopt a strategy that allows --moduleResolution node10 to work even without "exports" support.

shepherd.js/scripts/postprocess.js
joshkel: In case it helps, my experience using Shepherd.js 12.0.1, TypeScript 5.4, and Node.js 18.18.2:

* CommonJS, `"module": "NodeNext", "moduleResolution": "NodeNext"` (aka `Node16`) - works
* ESM, `"module": "NodeNext", "moduleResolution": "NodeNext"` (aka `Node16`) - compile error, "Property 'Tour' does not exist on type 'typeof import("node_modules/shepherd.js/dist/shepherd")'.
* CommonJS, `"module": "CommonJS", "moduleResolution": "Node"` (aka `Node10`) - compile error, "There are types at 'node_modules/shepherd.js/dist/shepherd.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'."

PR #2839, opened 2024-05-26
Export CSS and just create one file

RobbieTheWagner:
This goes back to one CSS file and adds it to exports so that folks should be able to import it.

### DIFF
--- a/docs-src/package.json
+++ b/docs-src/package.json
@@ -2,6 +2,7 @@
   "name": "shepherd-docs",
   "type": "module",
   "version": "0.0.6",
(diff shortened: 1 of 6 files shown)


Kenneth-Sills:
@renews The only thing in the way is that the linting workflow (specifically the ATTW step) is having trouble with the CSS, which Robbie has asked about [upstream](https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/30). I've answer himed there for visibility, but looks like it'll be a quick fix by excluding the CSS export from type checks.

PR #2853, opened 2024-06-04
fix: reconfigure CSS bundling

### LINKED ISSUE #2847: No longer able to exclude styles after upgrading from v12.0.4 to v12.0.5
Hi all,

After upgrading from v12.0.4 to v12.0.5, the default shepherd css styles are now included even without the use of:
`import 'shepherd.js/dist/css/shepherd.css';`

Unfortunately for our usecase where we were not looking to use the shepherd css stylesheet and instead using our own css, the recent change has caused our tour styles to break due to the high specificity of the svelte component css rules.

**expected behaviour**
I can choose whether or not I want to use the default shepherd styles by including the `shepherd.css` stylesheet.

**actual behaviour**
All shepherdjs svelte component styles are added after using `import Shepherd from 'shepherd.js'`

_shepherdjs svelte component styles included in head_
Screenshot 2024-06-02 at 08 03 56

Related discussion on issue #2834.

Kenneth-Sills:
As discussed in [the parent ticket](https://github.com/shepherd-pro/shepherd/issues/2847#issuecomment-2146650041), this prevents our CSS styles from being included for any project importing our JS module. The approach only differs in that I kept a separate build step to copy the ESM CSS to `dist/css`, as it retains compatibility with filepath imports and some bundler plugins that don't identify as either `import` or `require`.

I've added an E2E test group that gives us some basic monitoring for regressions in the CSS inclusion behavior and I've made sure to:

- Run all linters for this project.
- Run all E2E testing for this project.
- Clone vue-shepherd, [link](https://docs.npmjs.com/cli/v10/commands/npm-link) to this fixed branch, then run all E2E testing for that project.

The fix appears to work and no other regressions were seen.

If we want to be *extra* careful, we can ask the issue author to install from this branch and make sure it fixes their issue before release.

Closes #2847


RobbieTheWagner:
@Kenneth-Sills I am not a big fan of generating ESM and CJS CSS. I think our single CSS file we are currently generating should be fine. To my understand, I thought we just needed to change the `exports` somehow in package.json?

Kenneth-Sills:
See my explanation of what caused the regression [here](https://github.com/shepherd-pro/shepherd/issues/2847#issuecomment-2146650041). In summary, with how the Svelte Rollup plugin works, we *must* generate CSS for both builds or the CSS will be inserted in the JS and inserted at runtime.

Kenneth-Sills:
I've updated the build and confirmed via `pnpm run lint` and `pnpm run test:ci`.

Full disclosure though, I had my surgery today. So I'm on some pain medication and don't have access to the machine with the separate test project I was using. Subsequently, the commit changes are worth extra scrutiny and you *may* want to independently test it.

Thanks for your feedback and time!

RobbieTheWagner:
Thank you for all the hard work @Kenneth-Sills! I made some tweaks to reduce the number of builds, but this is great work.

RobbieTheWagner (approved):

chuckcarpenter:
Thank you and speedy recovery @Kenneth-Sills

PR #2870, opened 2024-06-13
✨ Add new package specific to Pro features

chuckcarpenter:


packages/pro-js/README
```diff
@@ -0,0 +1,5 @@
+# @shepherdpro/pro-js
```
RobbieTheWagner, line 1: Should this be `README.md`?

packages/pro-js/src/DataRequest.ts
```diff
@@ -12,9 +12,9 @@ interface TourStateDb extends DBSchema {
 }
 
 class DataRequest {
-  apiKey: string;
-  apiPath: string;
-  properties?: { [key: string]: unknown };
+  private apiKey: string;
+  private apiPath: string;
+  private properties?: { [key: string]: unknown };
```
RobbieTheWagner, line 17: The React package was complaining about these being private. Do they need to be? Wondering if using the JS way of making them private would make TS happier? Like `#apiKey` or whatever

RobbieTheWagner (approved):
More or less looks good to me 👍 . Left a couple comments.

packages/pro-js/src/DataRequest.ts
chuckcarpenter: interesting, I don't have the issue b/c the react package isn't reading them, just passing them through. I agree with following correct syntax though.

PR #2982, opened 2024-09-20
✨ Update shepherd element to use html dialog

### LINKED ISSUE #2959: Use the dialog element or the aria-modal attribute for the dialog modal of the tour tips
From the perspective of screenreader users the dialog modal tour tips in shephard have one problem. At the moment they are wrapped in a regular div element. The elements in the background of the highlighted component and the tour tip dialog modal when a tour is started are not clickable. Problem for screenreader users, the elements in the background are still accessible and are not excluded from the accessibility object model. for the screenreader user it is impossible to distinguish between what belongs to a dialog modal and what are elements in the background of the tour tip:  

https://github.com/user-attachments/assets/e3f35fcc-53fe-4820-a5ba-1f111c78ffa1

A way to tackle that would be making the switch from the `div` element to the `dialog` element for the dialog modals of the tour tips. the support is solid meanwhile and it is supported in all major browsers: https://caniuse.com/?search=dialog . there is also a highly recommended presentation outlining the ins and outs of the dialog element and the popover attribute.https://talks.hiddedevries.nl/VZNuWJ/dialog-dilemmas-and-modal-mischief-on-building-popovers#skMZ7WI

chuckcarpenter:
This issue addresses the request for semantic HTML and closes #2959.


shepherd.js/src/components/shepherd-element.svelte
```diff
@@ -141,20 +141,20 @@
   class:shepherd-element={true}
   {...dataStepId}
   on:keydown={handleKeyDown}
-  role="dialog"
-  tabindex="0"
```
chuckcarpenter, line 145: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#attributes

> Warning: The tabindex attribute must not be used on the <dialog> element.

shepherd.js/src/components/shepherd-element.svelte
```diff
@@ -141,20 +141,20 @@
...
   <ShepherdContent {descriptionId} {labelId} {step} />
-</div>
+</dialog>
 
 <style global>
   .shepherd-element {
     background: #fff;
     border-radius: 5px;
     box-shadow: 0 1px 4px rgba(0, 0, 0, 0.2);
+    margin: 0;
```
chuckcarpenter, line 157: margin auto is a default in the browser

test/cypress/integration/a11y.cy.js
```diff
@@ -75,7 +75,7 @@ describe('a11y', () => {
 
       cy.document().then(() => {
         cy.wait(1000);
-        cy.get('.shepherd-element').tab().tab().tab().tab().tab().tab();
+        cy.get('.shepherd-content').tab().tab();
```
RobbieTheWagner, line 78: Does this not work with all the previous tabs? This test is meant to tab a ton of times to show it is focus trapped in the modal

chuckcarpenter: It did not work, I can test some more though. It just didn't work as is, but now that I understand the point more I'll try again. Cannot be from the DIALOG though, so the content selector is needed.

RobbieTheWagner: We definitely want to make sure we keep the spirit of this test. It should tab at least like 4-6 times I think as it should select like the title, description, next button, previous button, etc and we want to make sure if you tab more times than the things inside that it loops around and stays focus trapped.

chuckcarpenter: We can pair and I can definitely show you that it does, so the correct way would be to make sure it loops around enough times to stay on the element with the data test attribute?

RobbieTheWagner: It just needs to prove that it loops around and does not start hitting elements outside of the tour

chuckcarpenter: finally got this working by adding some more to make sure it loops all the focus locked elements.

RobbieTheWagner (approved):
LGTM! Thanks for taking this on

PR #2995, opened 2024-10-04
Implement multiple element highlighting feature

### LINKED ISSUE #2487: Possible to separate highlighted div and placement of the tooltip
Hi,

I've come across a usecase that would make things a lot easier, and I don't know if it can be (easily) achieved or if it's indeed not possible.

I would like to highlight a div, but place the tooltip modal on another div (within the highlighted div). Maybe the following screenshot explains it a little bit better: 

image

Is this possible?

YuunsGit:
Hey 👋🏻

I'm in need of a feature that allows me to select multiple items in a single step. Currently, Shepherd highlights the overlay for the target element defined by the `attachTo.element` option.

To address this, I implemented an additional option, `extraHighlights`, in the `StepOptions`, along with an extra attribute in the `Step` object. While I saw [previous attempts](https://github.com/shipshapecode/shepherd/pull/2083) to add multi-highlight functionality (which suggested modifying the `attachTo` option to accept an array), I believe adding a separate option is more reasonable. 

Modifying the existing `attachTo` option is not ideal because it also handles the element targeting for the tooltip and the arrow placement. Refactoring this option to support an array would require reworking the entire logic for these aspects, which could introduce unnecessary complexity.

An example of the proposed approach using an array for `attachTo` would look something like this:
```js
attachTo: [
  {
    element: '.hero-including',
    on: 'bottom'
  },
  {
    element: '.pricing',
  },
  {
    element: '.docs',
  },
  {
    element: '#box-1',
  },
  {
    element: '#box-2',
  },
]
```
In this structure, there would only be a single `element` property for each object, which seems redundant.

Here's the current implementation with the new `extraHighlights` option:
```js
const step = new Step(tour, {
  attachTo: {
    element: '.hero-including',
    on: 'bottom'
  },
  extraHighlights: [ '.pricing', '.docs', '#box-1', '#box-2' ],
  ...moreOptions
});
```

Also closes #2487


YuunsGit:
Added `.replace(/\s/g, '')` to remove the redundant space from overlay path string returning from `makeOverlayPath`

Screenshot 2024-10-04 at 23 18 54

The unit tests should be fine now!

YuunsGit:
@RobbieTheWagner bump for notification
Sorry if bothered, it would be awesome to have this feature soon 🙂

RobbieTheWagner:
Hi @YuunsGit, thanks for the PR! Could you please add some test coverage for this?

YuunsGit:
> Hi @YuunsGit, thanks for the PR! Could you please add some test coverage for this?

Hi @RobbieTheWagner, sure thing! Implemented tests for the parts that I've modified or added.

YuunsGit:
I just noticed a mistake in the content of the step I added to the demo tour on the landing page, so I just removed my modifications on the landing page, thinking they weren't needed for this feature. Fyi, @RobbieTheWagner

YuunsGit:
Screenshot 2024-10-08 at 23 11 30

Also, @RobbieTheWagner, I just saw your comment on the other PR regarding this feature and that was such an edge case haha! I modified the code so that if an item to be highlighted is contained by another item it will be skipped.

image

And with this change this PR closes #2487

shepherd.js/src/step.ts
```diff
@@ -604,10 +632,17 @@ export class Step extends Evented {
...
     const content = this.shepherdElementComponent.getElement();
     const target = this.target || document.body;
+    const extraHighlightElements = this._resolvedExtraHighlightElements;
+
     target.classList.add(`${this.classPrefix}shepherd-enabled`);
     target.classList.add(`${this.classPrefix}shepherd-target`);
     content.classList.add('shepherd-enabled');
 
+    extraHighlightElements?.forEach((el) => {
+      el.classList.add(`${this.classPrefix}shepherd-enabled`);
```
RobbieTheWagner, line 642: Forgive my ignorance, but why do we want to add these classes to the extra highlighted elements?

RobbieTheWagner:
This is looking great @YuunsGit! I do think we should include an example in the demo and/or provide a cookbook example for this, so folks can find out more about it.

shepherd.js/src/step.ts
YuunsGit: I thought we would want the same classes applied to the target element (attackTo.element) on the extra highlighted elements as well.

RobbieTheWagner: I'm not sure if we do? We just want the cutout overlay I would assume? I am not sure what the classes do anymore though.

YuunsGit: One of the classes I added disables the click event if `canClickTarget`option is true

earlwilson:
@RobbieTheWagner Do you have a date planned for this to be rolled out?

shepherd.js/src/step.ts
```diff
@@ -184,6 +189,18 @@ export interface StepOptions {
...
+   * An array of extra element selectors to highlight when the overlay is shown
+   * The tooltip won't be fixed to these elements, but they will be highlighted
+   * just like the `attachTo` element.
+   * ```js
+   * const step = new Step(tour, {
+   *   extraHighlights: [ '.pricing', '#docs' ],
+   *   ...moreOptions
+   * });
+   */
+  extraHighlights?: ReadonlyArray<string>;
```
RobbieTheWagner, line 203: I think these are in alphabetical order. Would you mind adjusting placement of this please?

RobbieTheWagner (changes requested):

RobbieTheWagner:
> @RobbieTheWagner Do you have a date planned for this to be rolled out?

@earlwilson I will try to get it merged soon. Apologies for the delay!

shepherd.js/src/step.ts
YuunsGit: Oh I didn't notice that! Fixed, @RobbieTheWagner

RobbieTheWagner (approved):

RobbieTheWagner:
Thank you so much for the PR @YuunsGit! 🎉

PR #3011, opened 2024-10-11
Fix options type for `options.floatingUIOptions`

patrikholcak:
## About

Shepherd is already using `ComputePositionConfig` internally in a few places, but the options object itself is typed as `object`, which isn’t really helpful. Not sure if this is done because of some problem with docs — feel free to close this PR in that case!

### Proposed changes

*   add correct type to `Step.floatingUIOptions`

### DIFF
--- a/shepherd.js/src/step.ts
+++ b/shepherd.js/src/step.ts
@@ -17,6 +17,7 @@ import {
 // @ts-expect-error TODO: we don't have Svelte .d.ts files until we generate the dist
 import ShepherdElement from './components/shepherd-element.svelte';
 import { type Tour } from './tour.ts';
+import type { ComputePositionConfig } from '@floating-ui/dom';
 
 export type StepText =
   | string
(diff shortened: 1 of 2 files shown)


RobbieTheWagner (approved):

RobbieTheWagner:
@patrikholcak hitting this error:

```
17:38:26 [ERROR] [starlight-typedoc-plugin] node_modules/shepherd.js/src/utils/floating-ui.ts:63:17 - error TS2742: The inferred type of 'mergeTooltipConfig' cannot be named without a reference to '.pnpm/@floating-ui+core@1.6.5/node_modules/@floating-ui/core'. This is likely not portable. A type annotation is necessary.
```

patrikholcak:
Yeah, that explains why it hasn’t been added in the first place :smile: I don’t like doing it, but would you be ok with me copying some of the declarations from floating ui (bad, I know) or should I look into the docs plugin? Maybe there’s an option to ignore a type or make it "external"? :shrug:

I probably won’t have time over the weekend though

patrikholcak:
@RobbieTheWagner it should be fixed now (works on my machine!) let me know if it does and I’ll either cherry-pick this into #3012 or wait for this to get merged and then just rebase/merge these changes into #3012

leroy182 (commented):
@leroy182

PR #3012, opened 2024-10-11
Middleware deduping

patrikholcak:
## About

This might be a controversial (and breaking) change, but instead of overriding middlewares, I think the step middleware should always win. Feel free to close this PR or message me to make changes.

Take this example:

```
const defaultStepOptions = {
  floatingUIOptions: {
    middleware: [offset({ crossAxis: 32 })]
  }
};

const stepOptions = {
  floatingUIOptions: {
    middleware: [offset({ crossAxis: -32 })]
  }
}
```

To override `defaultStepOptions` you need equal negative spacing and sync these two numbers every time the default changes (or have a constant)

With this PR:

```
const defaultStepOptions = {
  floatingUIOptions: {
    middleware: [offset({ crossAxis: 32 })]
  }
};

const stepOptions = {
  floatingUIOptions: {
    middleware: [offset({ crossAxis: 0 })]
  }
}
```

The step offset middleware overrides the default step middleware

### DIFF
--- a/shepherd.js/src/step.ts
+++ b/shepherd.js/src/step.ts
@@ -8,7 +8,12 @@ import {
   isUndefined
 } from './utils/type-check.ts';
 import { bindAdvance } from './utils/bind.ts';
-import { parseAttachTo, normalizePrefix, uuid } from './utils/general.ts';
+import {
+  parseAttachTo,
+  normalizePrefix,
+  uuid,
+  dedupeMiddlewares
+} from './utils/general.ts';
 import {
   setupTooltip,
   destroyTooltip,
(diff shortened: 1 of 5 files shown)


chuckcarpenter:
@patrikholcak we have been talking about this recently and I recently released a change that changes the ordering, so that a dup added in would run last, but that still could have issues. This would be onus on the user to make sure they still have sane defaults when going down this path, but does give the most control.

chuckcarpenter:
@patrikholcak docs failed to build due to this error

image

patrikholcak:
Rebased and updated. I also updated a step test which was failing due to the changes in [dfb7d74](https://github.com/shipshapecode/shepherd/pull/3012/commits/dfb7d743d458e507f1d50b10972c62d1b4ab49ee). I’m not sure the test is relevant now, can probably be removed altogether :shrug:

PR #3051, opened 2024-11-08
Allow arrow padding to be configured for a step.

JakeThurman:
Hi! First time contributer.

We have a >4px border radius on out shepherd-element, and our looking to increase it so the arrow doesn't overlay the border radius:
image

This PR allows you to configure the arrow padding using something like: `arrow: { padding: 10 }`

This builds and tests pass locally, please let me know if there's anything else you need from me on this PR, or anything I missed!

### DIFF
--- a/shepherd.js/src/utils/floating-ui.ts
+++ b/shepherd.js/src/utils/floating-ui.ts
@@ -212,12 +212,15 @@ export function getFloatingUIOptions(
 
     if (arrowEl) {
       const arrowOptions =
-        typeof step.options.arrow === "object"
-        ? step.options.arrow
-        : { padding: 4 };
+        typeof step.options.arrow === 'object'
+          ? step.options.arrow
+          : { padding: 4 };
 
       options.middleware.push(
-        arrow({ element: arrowEl, padding: hasEdgeAlignment ? arrowOptions.padding : 0 })
+        arrow({
+          element: arrowEl,
+          padding: hasEdgeAlignment ? arrowOptions.padding : 0
+        })
       );
     }
 
--- a/shepherd.js/src/utils/floating-ui.ts
+++ b/shepherd.js/src/utils/floating-ui.ts
@@ -211,8 +211,13 @@ export function getFloatingUIOptions(
     );
 
     if (arrowEl) {
+      const arrowOptions =
+        typeof step.options.arrow === "object"
+        ? step.options.arrow
+        : { padding: 4 };
+
       options.middleware.push(
-        arrow({ element: arrowEl, padding: hasEdgeAlignment ? 4 : 0 })
+        arrow({ element: arrowEl, padding: hasEdgeAlignment ? arrowOptions.padding : 0 })
       );
     }
 
--- a/test/unit/components/shepherd-element.spec.js
+++ b/test/unit/components/shepherd-element.spec.js
@@ -44,6 +44,44 @@ describe('components/ShepherdElement', () => {
         container.querySelectorAll('.shepherd-element .shepherd-arrow').length
       ).toBe(0);
     });
+
+    it('arrow: object with padding shows arrow', async () => {
+      const testElement = document.createElement('div');
+      const tour = new Tour();
+      const step = new Step(tour, {
+        arrow: { padding: 10 },
+        attachTo: { element: testElement, on: 'top' }
+      });
+
+      const { container } = render(ShepherdElement, {
+        props: {
+          step
+        }
+      });
+
+      expect(
+        container.querySelectorAll('.shepherd-element .shepherd-arrow').length
+      ).toBe(1);
+    });
+    
+    it('arrow: empty object shows arrow', async () => {
+      const testElement = document.createElement('div');
+      const tour = new Tour();
+      const step = new Step(tour, {
+        arrow: {},
+        attachTo: { element: testElement, on: 'top' }
+      });
+
+      const { container } = render(ShepherdElement, {
+        props: {
+          step
+        }
+      });
+
+      expect(
+        container.querySelectorAll('.shepherd-element .shepherd-arrow').length
+      ).toBe(1);
+    });
   });
 
   describe('handleKeyDown', () => {
--- a/docs-src/src/content/docs/guides/usage.md
+++ b/docs-src/src/content/docs/guides/usage.md
@@ -297,4 +297,14 @@ myTour.addStep({
 });
 ```
 
+You can also provide an options object, to configure the arrow's [padding](https://floating-ui.com/docs/arrow#padding). The padding is the closest the arrow will get to the edge of the step.
+
+```js
+myTour.addStep({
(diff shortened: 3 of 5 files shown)


JakeThurman:
@RobbieTheWagner Any chance this could make it in soon? I'd rather not create an special fork for my company to use, and it sounds like this is important to have for us.

RobbieTheWagner:
Hi @JakeThurman apologies for the extremely late reply on this! Is this something that could be fixed via CSS in your app?

We do heavy arrow customization via CSS for our landing page with things like this:

```css
.shepherd-element[data-popper-placement^=top]>.shepherd-arrow {
  bottom:-8px
}
```

JakeThurman:
No worries on the delay. I appreciate you taking the time to reply now!

It's certainly possible, but it would only work in isolated cases. floating-ui has lots of logic to make sure the arrow points at the anchor element, and this would eliminate that down to just a fixed number of px from the edge. (i.e., middleware can shift the '
tour step 'balloon', but floating UI will adjust the arrow from there.)

This floating-ui padding parameter configures the closest the arrow can get to the edge of the balloon. It seems to be set to 4px because of the default shepherd CSS.

Finally, the floating-ui documentation explicitly says to set this parameter to your border-radius, which is what led me to open this PR so I could do so. See: https://floating-ui.com/docs/arrow#padding

shepherd.js/src/utils/floating-ui.ts
```diff
@@ -211,8 +211,13 @@ export function getFloatingUIOptions(
     );
 
     if (arrowEl) {
+      const arrowOptions =
+        typeof step.options.arrow === "object"
+        ? step.options.arrow
+        : { padding: 4 };
+
```
RobbieTheWagner, line 218:
```suggestion
```

shepherd.js/src/utils/floating-ui.ts
```diff
@@ -211,8 +211,13 @@ export function getFloatingUIOptions(
...
       options.middleware.push(
-        arrow({ element: arrowEl, padding: hasEdgeAlignment ? 4 : 0 })
+        arrow({ element: arrowEl, padding: hasEdgeAlignment ? arrowOptions.padding : 0 })
```
RobbieTheWagner, line 220:
```suggestion
        arrow({ element: arrowEl, padding: hasEdgeAlignment ? step.options.arrow?.padding ?? 4 : 0 })
```

What would you think about putting this inline here? I think it is a little less verbose.

test/unit/components/shepherd-element.spec.js
```diff
@@ -44,6 +44,44 @@ describe('components/ShepherdElement', () => {
         container.querySelectorAll('.shepherd-element .shepherd-arrow').length
       ).toBe(0);
     });
+
+    it('arrow: object with padding shows arrow', async () => {
+      const testElement = document.createElement('div');
+      const tour = new Tour();
+      const step = new Step(tour, {
+        arrow: { padding: 10 },
```
RobbieTheWagner, line 52: Thanks for adding tests! Could we perhaps add an assertion checking that padding of `10px` is applied to the arrow element?

test/unit/components/shepherd-element.spec.js
```diff
@@ -44,6 +44,44 @@ describe('components/ShepherdElement', () => {
...
+        attachTo: { element: testElement, on: 'top' }
+      });
+
+      const { container } = render(ShepherdElement, {
+        props: {
+          step
+        }
+      });
+
+      expect(
```
RobbieTheWagner, line 81: For this one, perhaps we could also assert padding, and I think it would be `0px` here?

RobbieTheWagner (changes requested):
This is great work, thank you for taking this on! I just had a few minor suggestions.

JakeThurman:
Pushed a fixup for the requested changes. I wasn't able to run my updated tests... 

I know I did originally when opening this PR, but I couldn't remember how to run them now. 
(I tried `pnpm run test:ci` didn't seem to work -- I got "No projects matched the filters in "C:\projects\shepherd")
Is that the right command?

RobbieTheWagner:
@JakeThurman apologies for the delay on this! The commands you will want to run tests are from the root: `pnpm test:cy:watch` or `pnpm test:unit:watch`

JakeThurman:
> test:cy:watch

Thanks! Updated my fixup to get the tests working.

RobbieTheWagner:
Thanks @JakeThurman I think things are looking good now! One final request, could you please update the docs to mention this? The auto generated docs will be updated based on your comments, but we have some hand rolled docs too like https://github.com/shipshapecode/shepherd/blob/main/docs-src/src/content/docs/guides/usage.md#-displaying-arrows

JakeThurman:
> Thanks @JakeThurman I think things are looking good now! One final request, could you please update the docs to mention this? The auto generated docs will be updated based on your comments, but we have some hand rolled docs too like [`main`/docs-src/src/content/docs/guides/usage.md#-displaying-arrows](https://github.com/shipshapecode/shepherd/blob/main/docs-src/src/content/docs/guides/usage.md#-displaying-arrows)

Done. I did a grep and didn't find any other documentation that appeared to need updating.

shepherd.js/src/utils/floating-ui.ts
```diff
@@ -211,8 +211,16 @@ export function getFloatingUIOptions(
     );
 
     if (arrowEl) {
+      const arrowOptions =
+        typeof step.options.arrow === 'object'
+          ? step.options.arrow
+          : { padding: 4 };
+
```
RobbieTheWagner, line 218:
```suggestion
```

shepherd.js/src/utils/floating-ui.ts
```diff
@@ -211,8 +211,16 @@ export function getFloatingUIOptions(
...
       options.middleware.push(
-        arrow({ element: arrowEl, padding: hasEdgeAlignment ? 4 : 0 })
+        arrow({
+          element: arrowEl,
+          padding: hasEdgeAlignment ? arrowOptions.padding : 0
```
RobbieTheWagner, line 222:
```suggestion
          padding: hasEdgeAlignment ? step.options.arrow?.padding ?? 4 : 0
```

RobbieTheWagner:
Thanks @JakeThurman! I think my suggestions might have gotten lost from before. I left a couple more, please let me know what you think of them.

shepherd.js/src/utils/floating-ui.ts
JakeThurman: I think this would be working javascript (I.e., `(true).padding` is `undefined`) but typescript doesn't allow that.

RobbieTheWagner: Oh, gotcha. Didn't realize TS wouldn't allow it.

RobbieTheWagner (approved):

PR #3068, opened 2024-12-06
💄 Add reset for dialog element

chuckcarpenter:


shepherd.js/src/components/shepherd-element.svelte
```diff
@@ -150,6 +150,18 @@
 </dialog>
 
 <style global>
+  /* <dialog> reset needed for positioning */
+  dialog.shepherd-element {
```
RobbieTheWagner, line 154: I think we could put just a handful of these properties on the `.shepherd-element` selector we already have and achieve what we are looking for. I do wonder if we should forcibly include some of this, rather than bundling in our CSS file though.

RobbieTheWagner (approved):

PR #3083, opened 2024-12-31
♿ Remove tabindex to reduce tabs for modal trap

### LINKED ISSUE #3071: Accessibility issue: unnecessary tabstop
This is tangentially related to #3065 and relates to focus management

Per the [W3C's APG](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/): "Generally, focus is initially set on the first focusable element. However, the most appropriate focus placement will depend on the nature and size of the content"

So, as the discussion in #3065 mentions, the Close button would be the more appropriate location for focus to be sent.

Another issue, which I'm raising, is that there's an inner `div` with `tabindex="0"`.  That `tabindex` should be removed -  firstly because the only things that are given a `tabindex` are things which are actionable, and secondly because it causes an unnecessary 2nd announcement of the dialog.

https://us06web.zoom.us/clips/share/A2F3MRZqZXlld0NNQ1NVaTdjbTFCZ2lIQXBBAQ

All-in-all, I want to thank you for creating this. It is really good and does mostly a great job with accessibility, which is very important to me.

chuckcarpenter:
closes #3071 

This removes the tabindex added by the content div that's not needed for tab trapping in the dialog.

### DIFF
--- a/shepherd.js/src/components/shepherd-content.svelte
+++ b/shepherd.js/src/components/shepherd-content.svelte
@@ -7,7 +7,7 @@
   export let descriptionId, labelId, step;
 </script>
 
-<div class="shepherd-content" tabindex="0">
+<div class="shepherd-content">
   {#if !isUndefined(step.options.title) || (step.options.cancelIcon && step.options.cancelIcon.enabled)}
     <ShepherdHeader {labelId} {step} />
   {/if}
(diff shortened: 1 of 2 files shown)


RobbieTheWagner (approved):
Nice, good find!

PR #3144, opened 2025-03-06
Svelte 5 that builds

RobbieTheWagner:


shepherd.js/src/step.ts
```diff
@@ -19,10 +19,10 @@ import {
   destroyTooltip,
   mergeTooltipConfig
 } from './utils/floating-ui.ts';
-// @ts-expect-error TODO: we don't have Svelte .d.ts files until we generate the dist
 import ShepherdElement from './components/shepherd-element.svelte';
 import { type Tour } from './tour.ts';
 import type { ComputePositionConfig } from '@floating-ui/dom';
+import { createClassComponent } from 'svelte/legacy';
```
RobbieTheWagner, line 25: @Rich-Harris is there a way to do this without the legacy stuff? We need to get a reference to an element, which does not seem to be possible in Svelte 5 unless the parent is also a Svelte component.

PR #3172, opened 2025-04-14
fix: Update install command

alph486:
This change just puts the install command in line with what the npm repository says.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -11,7 +11,7 @@ that can be passed into props for dynamic interactivity.
 Use this simple NPM command or whatever package manager is your favorite.
 
 ```bash
-npm install --save react-shepherd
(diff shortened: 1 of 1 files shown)


RobbieTheWagner:
@alph486 it should actually be `react-shepherd`. Could you please let me know where you are seeing otherwise?

alph486:
> @alph486 it should actually be `react-shepherd`. Could you please let me know where you are seeing otherwise?

Sure! When I click the npm badge on the readme.md for the react component it takes me here https://www.npmjs.com/package/@shepherdpro/react which indicates a different install command.

RobbieTheWagner:
@alph486 gotcha, I think this PR should instead update any of those references to point to `react-shepherd` as that is the correct one.

PR #3303, opened 2025-11-08
Use CSS anchor positioning

RobbieTheWagner:


RobbieTheWagner:
CSS anchor positioning is not going to be viable for our use case, so closing this PR

PR #3441, opened 2026-05-13
Fix mobile input focus loss during tooltip repositioning

### LINKED ISSUE #1143: Targeted input loses focus on android and ios devices
Hey, I am having a weird issue in mobile devices that I cannot figure out.

Basically I have one step that targets an input element, and it is expected that the user types something to proceed to the next step.

This works fine when trying on desktop, but on android or ios when I click the input element to start typing the input immediately loses focus. On android is very obvious that the keyboard shows for a little bit and disappears

Here's a simple codesanbox showing the issue: 
https://codesandbox.io/s/silly-ritchie-o0w88?file=/src/index.js

The issue doesn't appear in the browser's responsive mode, only in the actual devices.

One thing that I noticed is that If I remove the 'shepherd-element' element from the dom it works fine. Not sure if this indicates something relevant

Thanks

serendipitous-syntax:
Fixes #1143.

This keeps the existing focus-after-render behavior for the initial step render, but skips the delayed `step.el.focus()` on later Floating UI auto-update/reposition calls. That prevents mobile viewport/keyboard resize updates from stealing focus back from an input attached to the active step.

The regression test covers the focus behavior directly: the step element is focused once after initial render, then a focused input remains focused after a later reposition update.

Checks run:
- CYPRESS_INSTALL_BINARY=0 npx pnpm install --lockfile=false
- npx pnpm -F shepherd.js exec vitest run test/unit/utils/floating-ui.spec.js
- npx pnpm -F shepherd.js test:unit:ci
- npx pnpm -F shepherd.js lint:js
- npx pnpm -F shepherd.js lint:prettier
- npx pnpm -F shepherd.js types:check
- npx pnpm -F shepherd.js build
- git diff --check

Notes:
- `CYPRESS_INSTALL_BINARY=0 npx pnpm install --frozen-lockfile` is currently blocked by a duplicate `@vitest/pretty-format@4.1.5` mapping in `pnpm-lock.yaml`.
- Cypress browser tests were not run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip positioning across centered and non-centered steps.
  * Prevented tooltips from repeatedly stealing focus after initial display.
  * Improved handling of custom arrow and positioning configurations.

### DIFF
--- a/shepherd.js/src/utils/floating-ui.ts
+++ b/shepherd.js/src/utils/floating-ui.ts
@@ -33,6 +33,8 @@ export function setupTooltip(step: Step): ComputePositionConfig {
   let target = attachToOptions.element as HTMLElement;
   const floatingUIOptions = getFloatingUIOptions(attachToOptions, step);
   const shouldCenter = shouldCenterStep(attachToOptions);
+  // Keep this function-local so every fresh step render gets one focus attempt.
+  let shouldFocusAfterRender = true;
 
   if (shouldCenter) {
     target = document.body;
@@ -47,7 +49,10 @@ export function setupTooltip(step: Step): ComputePositionConfig {
       return;
     }
 
-    setPosition(target, step, floatingUIOptions, shouldCenter);
+    setPosition(target, step, floatingUIOptions, shouldCenter, {
+      shouldFocusAfterRender
+    });
+    shouldFocusAfterRender = false;
   });
 
   step.target = attachToOptions.element as HTMLElement;
@@ -92,11 +97,21 @@ function setPosition(
   target: HTMLElement,
   step: Step,
   floatingUIOptions: ComputePositionConfig,
-  shouldCenter: boolean
+  shouldCenter: boolean,
+  { shouldFocusAfterRender }: { shouldFocusAfterRender: boolean }
 ) {
+  const positionPromise = computePosition(
+    target,
+    step.el as HTMLElement,
+    floatingUIOptions
+  ).then(floatingUIposition(step, shouldCenter));
+
+  if (!shouldFocusAfterRender) {
+    return positionPromise;
+  }
+
   return (
-    computePosition(target, step.el as HTMLElement, floatingUIOptions)
-      .then(floatingUIposition(step, shouldCenter))
+    positionPromise
       // Wait before forcing focus.
       .then(
         (step: Step) =>
--- a/shepherd.js/test/unit/utils/floating-ui.spec.js
+++ b/shepherd.js/test/unit/utils/floating-ui.spec.js
@@ -1,7 +1,27 @@
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const floatingUIMock = vi.hoisted(() => ({
+  autoUpdate: vi.fn(),
+  computePosition: vi.fn(),
+  updateCallbacks: []
+}));
+
+vi.mock('@floating-ui/dom', async (importOriginal) => {
+  const actual = await importOriginal();
+
+  return {
+    ...actual,
+    autoUpdate: floatingUIMock.autoUpdate,
+    computePosition: floatingUIMock.computePosition
+  };
+});
+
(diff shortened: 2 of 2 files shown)


chuckcarpenter:
Thanks for this, @serendipitous-syntax, and sorry it sat unreviewed for three months. I dug into it properly — the core change is correct, and I want to lay out what I verified plus the two things standing between it and a merge.

### The fix is right, and I checked the part I was most worried about

My concern going in was scope: if `shouldFocusAfterRender` were module-scoped, only the first step of a tour would ever receive focus and every subsequent step would silently lose keyboard and screen-reader users — worse than the bug being fixed.

It isn't. `let shouldFocusAfterRender = true` is a function-local in `setupTooltip`, captured by the `autoUpdate` closure, and `setupTooltip` has exactly one call site: `Step._setupElements()`, which `_show()` calls on every show after `_teardownElements()` nulls `this.el`. New element, new closure, flag back to `true`. So it's per-step-**show**, not per-tour.

Verified rather than assumed, in real headless Chrome against a three-step tour:

```
step1 -> DIALOG.shepherd-element[step=s1]
step2 -> DIALOG.shepherd-element[step=s2]
step3 -> DIALOG.shepherd-element[step=s3]
back() -> DIALOG.shepherd-element[step=s2]
hide()+show() -> DIALOG.shepherd-element[step=s2]
tour A complete, tour B start -> DIALOG.shepherd-element[x1]   (no cross-tour leak)
```

All 11 existing `a11y.cy.js` tests pass, and a probe confirming a post-first-render viewport resize no longer steals focus also passes. Applying just your `floating-ui.ts` onto current `main`: 225 unit tests green, `eslint` clean, `types:check` clean.

### Blocker 1: the `eslint-plugin-cypress` pin breaks CI

Commit `2a23169` pins `eslint-plugin-cypress` to `5.3.0` in `shepherd.js/package.json`, but `pnpm-lock.yaml` still records `^5.3.0`. `.github/workflows/test.yml` runs a bare `pnpm install`, and pnpm defaults to `--frozen-lockfile` when `CI=true` — so this fails with `ERR_PNPM_OUTDATED_LOCKFILE` at the install step of **all three** jobs, turning a one-job failure into a three-job failure.

It also isn't needed: `eslint .` exits 0 on current `main` with your source change and no pin. And it's the sole cause of the current merge conflict — `floating-ui.ts` merges cleanly.

Worth knowing: the frozen-install failure you hit (duplicate `@vitest/pretty-format@4.1.5`) was real at your branch point but is already gone on `main`, so a rebase clears it without the pin.

Could you drop that commit and rebase onto `main`, keeping only `63b7bef`?

### Blocker 2: CI has never actually run on this branch

`actions/runs?head_sha=2a23169` returns `total_count: 0`. The green-looking checks are CodeRabbit plus a `qlty` run that reports "Skipping analysis of pull request from a fork". The one real run was against your first commit and failed lint for a reason whose logs have since expired. So there's no CI evidence either way yet — the rebase should fix that.

### One request before merge

The regression test doesn't guard the invariant it exists to protect. I hoisted `shouldFocusAfterRender` to module scope — exactly the refactor that should trip it — and the spec still passed 7/7, while focus broke on every step from 2 onward and on every subsequent tour. The Cypress suite misses it too; `a11y.cy.js` only asserts focus at tour *start*.

Could you add a case that calls `setupTooltip` twice with a fresh element the second time (simulating step 2 / `tour.back()`) and asserts focus fires once per call? That's the thing a future refactor would break.

### Optional, not blocking

- Consider clearing the flag only once focus actually lands (checking `document.activeElement === step.el` in the final `.then`). `main` got an implicit retry for free from repositioning; there's now exactly one attempt per show. I couldn't construct a case where the single attempt fails, so this is hardening rather than a defect.
- A named parameter instead of the 5th positional boolean on `setPosition`, plus a comment noting the flag is per-step-show — that's the one non-obvious fact a future reader will get wrong.

For anyone finding this later: this fixes #1143. It does **not** close #2117 (still no `autoFocus: false` opt-out — first-render focus is preserved by design) and it neither fixes nor breaks #1793.

Nice piece of debugging — the diagnosis in the original 2020 thread was right and this is the correct mechanism for it.

PR #3493, opened 2026-08-25
feat: serve markdown from the homepage via Accept negotiation

RobbieTheWagner:
Part 3 of 6 of the agent-readiness stack (stacked on #3492).

The homepage now renders on demand (`prerender = false`, using the same serverless infra as `/api/checkout`) so it can content-negotiate per [acceptmarkdown.com](https://acceptmarkdown.com): requests with `Accept: text/markdown` get a markdown rendition with `Content-Type: text/markdown` and `Vary: Accept`; HTML responses also send `Vary: Accept`. The markdown is additionally served at `/index.md`, and the sitemap gains `customPages` for the no-longer-prerendered homepage.

Why on-demand rendering: `vercel.json` rewrites are unreliable with the Astro adapter, and Vercel edge middleware does not run for prerendered pages — this is the only mechanism guaranteed to work. The response is uncached (`max-age=0, must-revalidate`) so a CDN can never serve the wrong variant; the tradeoff is a function invocation per homepage hit.

**Test plan:** 9 unit tests for the Accept q-value parser, e2e tests for both negotiation directions + `/index.md`, and build-output assertions (homepage in sitemap, `index.md` emitted, no prerendered `index.html`) — 21 tests passing on this branch. Verified in the browser that the homepage and demo tour behave identically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added Markdown content negotiation for the homepage based on the `Accept` header.
  - Added a dedicated `/index.md` endpoint with homepage content, installation instructions, usage examples, licensing, and links.
  - Homepage responses now correctly advertise content variation and support on-demand rendering.
  - Updated sitemap output to include the production homepage.

- **Bug Fixes**
  - Improved handling of browser preferences, wildcards, quality values, and explicit refusals when selecting Markdown or HTML.

- **Tests**
  - Added coverage for content negotiation, Markdown responses, sitemap output, and generated landing-page files.

### DIFF
--- a/landing/astro.config.mjs
+++ b/landing/astro.config.mjs
@@ -12,7 +12,15 @@ export default defineConfig({
   // sitemap need to use www to avoid redirect chains.
   site: 'https://www.shepherdjs.dev',
 
-  integrations: [mdx(), sitemap()],
+  integrations: [
+    mdx(),
+    sitemap({
+      // The homepage is rendered on demand (for markdown content
+      // negotiation), so the sitemap integration cannot discover it
+      // at build time.
+      customPages: ['https://www.shepherdjs.dev/']
+    })
+  ],
 
   output: 'static',
   adapter: vercel(),
--- a/landing/src/lib/accept.ts
+++ b/landing/src/lib/accept.ts
@@ -1,0 +1,78 @@
+interface MediaRange {
+  type: string;
+  subtype: string;
+  q: number;
+  specificity: number;
+}
+
+function parseAccept(accept: string): MediaRange[] {
+  return accept
+    .split(',')
+    .map((part) => part.trim())
+    .filter(Boolean)
+    .map((part) => {
+      const [range, ...params] = part.split(';');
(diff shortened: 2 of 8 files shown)


RobbieTheWagner:
Good catch — the preview was serving raw markdown to everyone. Root cause: with the homepage no longer prerendered there is no static `index.html`, and Vercel's filesystem handler (which runs **before** the `^/$ → _render` route) fell back to resolving `/` to the prerendered static `index.md` as the directory index, so the negotiation function never ran.

Fixed in c39faa73: `/index.md` is now rendered on demand as well, so the static output contains no root index files and `/` always reaches the negotiation route. The build-output tests now assert this can't regress (no static `index.html`/`index.md`, and both `/` and `/index.md` route to the render function in the deployment config).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

